PR #6, opened 2017-08-03
Prevent chart methods from being called after the chart is destroyed

AlexMayants:
I noticed that if the `HighchartsChart` component is unmounted as a whole from a page, exceptions get thrown from inside the Highcharts code. This disrupts the unmounting process and prevents transitions to another page if react-router is used.

The exceptions happen because the `chart.destroy` method inside the `BaseChart` gets called before unmounting Axis and Series components which call Highcharts methods to do their cleanup.

In this pull request I attempt to fix this. I haven't found a way to change the order of unmounting, so I wrapped the chart methods passed via providers in a wrapper which checks whether the chart is already destroyed.


packages/react-jsx-highcharts/src/components/Axis/Axis.js
```diff
@@ -39,7 +39,9 @@ class Axis extends Component {
   }
 
   componentWillUnmount () {
-    this.props.remove();
+    const { remove } = this.props;
+
+    remove && remove();
```
whawker: Is this still needed with the `getBoundChartMethod` util method?

packages/react-jsx-highcharts/src/components/AxisProvider/index.js
```diff
@@ -29,15 +30,16 @@ export default function provideAxis(WrappedComponent) {
       const id = this.props.axisId || this.props.id;
       if (!id) return null;
 
+      const chart = this.props.getChart();
       const axis = this.props.get(id);
```
whawker: Just to remove a bit of duplication, could we add a helper function here?

```js
const boundAxisHelper = (chart, axis) => method => {
  return getBoundChartMethod(chart, method, axis);
}
const getBoundAxisMethod = (this.props.getChart(), axis);

const update = axis && getBoundAxisMethod(axis.update)
// etc
```

packages/react-jsx-highcharts/src/components/ChartProvider/index.js
```diff
@@ -34,10 +35,10 @@ export default function provideChart(WrappedComponent) {
...
           {...this.props}
           get={chart.get.bind(chart)}
-          update={cleanPropsBeforeUpdate(chart.update.bind(chart))}
-          addAxis={cleanPropsBeforeUpdate(chart.addAxis.bind(chart))}
-          addSeries={cleanPropsBeforeUpdate(chart.addSeries.bind(chart))}
-          setTitle={cleanPropsBeforeUpdate(chart.setTitle.bind(chart))}
+          update={cleanPropsBeforeUpdate(getBoundChartMethod(chart, chart.update, chart))}
+          addAxis={cleanPropsBeforeUpdate(getBoundChartMethod(chart, chart.addAxis, chart))}
+          addSeries={cleanPropsBeforeUpdate(getBoundChartMethod(chart, chart.addSeries, chart))}
+          setTitle={cleanPropsBeforeUpdate(getBoundChartMethod(chart, chart.setTitle, chart))}
```
whawker: Ditto with adding a helper method, as mentioned on the `AxisProvider`

packages/react-jsx-highcharts/src/components/SeriesProvider/index.js
```diff
@@ -52,10 +53,11 @@ export default function provideSeries(WrappedComponent, expectsSeriesExists = tr
...
 
-      const update = series && series.update.bind(series);
-      const remove = series && series.remove.bind(series);
-      const setData = series && series.setData.bind(series);
-      const setVisible = series && series.setVisible.bind(series);
+      const chart = this.props.getChart();
+      const update = series && getBoundChartMethod(chart, series.update, series);
+      const remove = series && getBoundChartMethod(chart, series.remove, series);
+      const setData = series && getBoundChartMethod(chart, series.setData, series);
+      const setVisible = series && getBoundChartMethod(chart, series.setVisible, series);
```
whawker: Ditto with adding a helper method, as mentioned on the `AxisProvider`

packages/react-jsx-highcharts/src/utils/getBoundChartMethod.js
```diff
@@ -0,0 +1,11 @@
+export default function getBoundChartMethod(chart, method, context) {
+  const boundMethod = method.bind(context);
+
+  return function (...rest) {
```
whawker: Nitpicking, but could we use the convention `...args` here, instead of `...rest`?

`...rest` implies we've destructured some other arguments.

packages/react-jsx-highcharts/src/utils/getBoundChartMethod.js
```diff
@@ -0,0 +1,11 @@
...
+    if (!chart.__destroyed) {
+      return boundMethod(...rest)
+    }
+
+    return null;
```
whawker: P.s. I wonder if we returned a noop function (`noop = () => {}`) here instead, we might not need to

```js
componentWillUnmount () {
  const { remove } = this.props;

  remove && remove();
}
```

We could just call `this.props.remove()` as before - it just wouldn't do anything.

packages/react-jsx-highcharts/src/components/Axis/Axis.js
AlexMayants: I added this check here and in Series.js, because Providers do not return methods when parent objects don't exist (i.e. `axis && axis.remove`), and `componentWillUnmount` could be called after the axis was destroyed with the chart.

I upgraded the `getBoundChartMethod` to return `noop` when method is empty, so this code isn't need any more.

packages/react-jsx-highcharts/src/components/AxisProvider/index.js
AlexMayants: added a `boundContextHelper` to `getBoundChartMethod`. The code got much cleaner indeed, thanks =)

packages/react-jsx-highcharts/src/utils/getBoundChartMethod.js
AlexMayants: Sure, why not =)

packages/react-jsx-highcharts/src/utils/getBoundChartMethod.js
AlexMayants: As I mentioned above, this construction isn't due to `return null` here. The `noop` is a nice idea though, so `getBoundChartMethod` now encapsulates `context && context.method` part from providers.

AlexMayants:
Thanks for your suggestions! I made some changes to the PR and cleaned up the `examples` dir.

whawker:
Amazing, thanks @AlexMayants! I'll merge now and add you as a contributor.

whawker:
Just published this as `react-jsx-highcharts` and `react-jsx-highstock` 1.1.0

PR #7, opened 2017-08-08
Do not fail during initalization when Highcharts is not instantiated

AlexMayants:
Our application uses universal rendering. It turns out, if there is no `window` in the current environment (such as server rendering), the `'highcharts'` dependency returns a factory function, not the Highcharts namespace instance. Because of this, a failure happens at [Tooltip.js#L16](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js#L16).

This pull request attempts to solve the problem by checking whether the `Highcharts.defaultOptions` do exist when initializing `Tooltip`'s `defaultProps`, so the application doesn't fail on class definition. Usage of `Highcharts` inside methods (`Highcharts.addEvent()`, `Highcharts.chart()` etc.) needs further research though.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
+++ b/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
@@ -13,7 +13,7 @@ class Tooltip extends Component {
   };
 
   static defaultProps = {
-    ...Highcharts.defaultOptions.tooltip,
+    ...(Highcharts.defaultOptions && Highcharts.defaultOptions.tooltip),
     enabled: true
   };
 
--- a/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js
+++ b/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js
@@ -13,7 +13,7 @@ class RangeSelector extends Component {
   };
 
   static defaultProps = {
-    ...Highcharts.defaultOptions.rangeSelector,
+    ...(Highcharts.defaultOptions && Highcharts.defaultOptions.rangeSelector),
     enabled: true
   };
 


packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
```diff
@@ -13,7 +13,7 @@ class Tooltip extends Component {
   };
 
   static defaultProps = {
-    ...Highcharts.defaultOptions.tooltip,
+    ...(Highcharts.defaultOptions && Highcharts.defaultOptions.tooltip),
```
whawker: Would you mind applying the same fix to [RangeSelector](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js#L16)?

whawker: P.s. Does `Highcharts.getOptions().tooltip` work by any chance?

whawker:
In the application you're working on publicly accessible? I would love to add it to a "real-world" example if possible.

packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
AlexMayants: Fixed the [RangeSelector](/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js#L16) too.

I haven't tested `Highcharts.getOptions()`, but since this method is defined inside the same factory function, it probably won't work.

AlexMayants:
Added the fix to [RangeSelector](/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js#L16) too.

> Is the application you're working on publicly accessible?

Not yet, it's still in development. I'll give you a link when we deploy it, though.

PR #25, opened 2017-09-26
Allow className property on HighchartsChart

will-stone:


packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
```diff
@@ -85,7 +85,7 @@ class BaseChart extends Component {
   render () {
     return (
       <div
-        className="chart"
+        className={`chart ${this.props.className}`}
```
whawker: Thanks for the PR! 😄 

Mind adding a default empty string prop for `className`? Otherwise we'll end up having `class="chart undefined"`.

I'd do it myself but I only have my phone at the moment!

will-stone:
@whawker I've made the changes and added a small function around the class names that means that if className isn't provided you won't end up with a dangling space e.g. `class="chart "`

I usually use the `classnames` package but I thought you wouldn't want another dependency.

whawker:
Thanks!!

PR #92, opened 2018-04-09
Build ES6 module for highcharts

anajavi:
This PR adds ES module build for react-jsx-highcharts.
This lets webpack 4 do tree shaking and module concatenation which saves me around 50kB in final bundle.

As I am not too familiar with the source code, I am not 100% sure that setting "sideEffects": false is entirely safe. On the surface it seems to be.
The sideEffects flag can also be array of globs if there are side effects of importing some modules.

It also modifies the ES build to use lodash-es instead of lodash. That babel plugin is copied from https://github.com/PaulLeCam/react-leaflet
The lodash-es allows some more tree shaking.
As there is no commonjs build, the lodash-es could be used in unmodified source too.

I also tried building UMD with webpack 4 and that also saves some space. I can make separate PR for it if you like?

I did not add this build to highstock package as I don't have real world application to test it with.

Do you think this is okay to be included in react-jsx-highcharts?

### DIFF
--- a/packages/react-jsx-highcharts/.babelrc
+++ b/packages/react-jsx-highcharts/.babelrc
@@ -12,6 +12,9 @@
     },
     "production": {
       "plugins": ["lodash", "transform-runtime", "transform-react-remove-prop-types"]
+    },
+    "es": {
+      "plugins": ["./build/use-lodash-es", "lodash", "transform-runtime", "transform-react-remove-prop-types"]
     }
   }
 }
--- a/packages/react-jsx-highcharts/build/use-lodash-es.js
+++ b/packages/react-jsx-highcharts/build/use-lodash-es.js
@@ -1,0 +1,10 @@
+module.exports = function() {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        var source = path.node.source
+        source.value = source.value.replace(/^lodash($|\/)/, 'lodash-es$1')
+      },
+    },
+  }
+}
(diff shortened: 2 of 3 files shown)


whawker:
Wow amazing! Thanks for this! 

I think we should probably update the devDependencies to use webpack 4 too, as the build here specifically relates to that?

anajavi:
Webpack 4 is not needed for ES build, but it makes smaller UMD build. I will send another PR for it.

I was pretty amazed by 50kB saving in real world application. Most of it comes from reusing lodash-es by react-leaflet and our own code. Without lodash-es the saving was 20kB.

anajavi:
The babel plugin transform-runtime seems to add dependency to babel-runtime and core-js.

Why is the transform-runtime needed? It might be possible to disable it for the es-build rather than declare dependencies in package.json.

whawker:
Looks good to me! Merging!

anajavi:
Just to remind that babel-runtime needs to be added to dependencies.

whawker:
Ah thanks, devDependencies I assume?

anajavi:
In dependencies or peerDependencies.

ES6 is not bundled, so everything that is used in code must be in dependencies or peerDependencies.

I forgot babel-runtime, because I had it on my own project anyway.

anajavi:
If you like, I can test the final package, if you publish it with tag to npm?

whawker:
Yeah just testing the webpack 4 stuff, then I'll push it out as an alpha release.

whawker:
@anajavi published as `2.3.0-alpha.1`

whawker:
No no, thank **you**!

PR #93, opened 2018-04-09
Upgrade react-jsx-highcharts build to webpack 4

anajavi:
Upgrade webpack to version 4.

Saves some 10kB on minified build.


anajavi:
Using lodash-es instead of lodash in UMD build saves another 2kB.

Might get half kB with making is-immutable es module too.

PR #95, opened 2018-04-14
Cleanup babelrc

anajavi:
Removes stage-0 preset and cleans up babelrc.

This also rewrites index.js exports to standard, instead of the currently stage-1 proposal. It makes it easier to use src folder directly in npm package.


whawker:
Thanks so for all these PRs! I've been hoping to have some help maintaining going forward.

Unfortunately however, I'm on holiday at the moment and don't have my laptop with me, I won't be able to test the PRs till I'm back on the 18th. Hope that doesn't block work on your end!

anajavi:
No worries. I'm not in a hurry. Have a fun holiday!

PR #100, opened 2018-05-12
Fix uuid uniqueness

anajavi:
Makes Axis and Series uuid's unique for every instance.

Also drops deprecated componentWillMount from those classes. Options3d I did not know how to test, although removing it from there looks straight forward.

This might not be the cleanest approach, and it might have bugs..
Feel free to ignore if you come up with some better way.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Axis/Axis.js
+++ b/packages/react-jsx-highcharts/src/components/Axis/Axis.js
@@ -11,7 +11,7 @@ class Axis extends Component {
 
   static propTypes = {
     type: validAxisTypes,
-    id: PropTypes.string.isRequired,
+    id: PropTypes.string,
     children: PropTypes.node,
     getChart: PropTypes.func, // Provided by ChartProvider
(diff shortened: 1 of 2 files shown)


whawker:
I'm thinking of closing this in favour of using `getDerivedStateFromProps` with the [react-lifecycles-compat polyfill](https://www.npmjs.com/package/react-lifecycles-compat). Previously I was unaware this polyfill existed.

That way we don't need to do `this.forceUpdate()`. Any thoughts?

whawker:
Ahh, now I see the purpose of `forceUpdate`, because otherwise the provider will not get re-rendered when `this.axis` is set?

anajavi:
Yes, thats the purpose. Sorry I didn't add comment line there.

Somehow, in Series it got re-rendered without forceUpdate, but as that might be timing related I added it there also.

whawker:
In that case, I will merge. I think it's unavoidable with deprecated methods. Thanks for this!

PR #101, opened 2018-05-13
Remove default xaxis id

anajavi:
Removes the xAxis id default value. This makes it possible to add multiple XAxises without providing id's.

Hope it doesn't have any special meaning somewhere else.


whawker:
I think the only place this is useful is for the [highstocks-datepickers](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highstock-datepickers/README.md). But it would just be a readme update.

PR #103, opened 2018-05-14
Upgrade mocha & lodash

anajavi:
Running npm audit had warnings about mocha and lodash:
https://nodesecurity.io/advisories/534 
https://nodesecurity.io/advisories/146 
https://nodesecurity.io/advisories/577 

I don't know if these are at all relevant with this library, but it shows good style to upgrade them:)


whawker:
Good point! Thanks!

PR #106, opened 2018-05-22
Update datepickers build to webpack 4

anajavi:
This updates datepickers to use webpack 4. Doesn't save much space, but it is more consistent with other modules.

I didn't test this at all, because I have never used it. Hope it still helps a little to release 3.0.0.


whawker:
Legend, thanks so much @anajavi!

PR #116, opened 2018-06-07
Add flow typings for external API

anajavi:
This adds flow typings for library consumers.

build and test scripts are modified:
- build:prod builds all targets now
- test runs mocha and tests flow types with test file under /test/flow-types

I'm pretty sure there are some problems, but I'd like somebody else to try it out as it is easier to find bugs that way.

Simplest way to try it out:
- edit test/flow-types/index.js.flow
- npm run test:flow

### DIFF
--- a/packages/react-jsx-highcharts/src/index.js.flow
+++ b/packages/react-jsx-highcharts/src/index.js.flow
@@ -1,0 +1,182 @@
+// @flow
+import React from 'react';
+import type { Node, ComponentType, ElementConfig } from 'react';
+import type HighchartsType from 'highcharts';
+
+
+type HighchartsChartProps = {
+  children?: Node
+}
+
+export function HighchartsChart(props: HighchartsChartProps) { }
+
+type HighchartsSparklineProps = {
+  height?: number,
+  width?: number,
+  margin?: Array<number>,
+  style?: {},
+  plotOptions?: {},
+  series?: Node,// note series or children required
+} & HighchartsChartProps
+
+export class HighchartsSparkline extends React.Component<HighchartsSparklineProps> { }
+
+
+type Highcharts3dChartProps = {
+  alpha?: number,
+  axisLabelPosition?: string,
+  beta?: number,
+  depth?: number,
+  fitToPlot?: boolean,
+  frame?: {},
+  viewDistance?: number,
+} & HighchartsChartProps;
+
+export class Highcharts3dChart extends React.Component<Highcharts3dChartProps> {
+
+}
+
+type ChartProps = {
+  type?: string,
+  width?: number,
+  height?: number,
+  onAddSeries?: Function,
+  onAfterPrint?: Function,
+  onBeforePrint?: Function,
+  onClick?: Function,
+  onLoad?: Function,
+  onRedraw?: Function,
+  onRender?: Function,
+  onSelection?: Function,
+}
+export function Chart (props: ChartProps) {}
+type CreditsProps = {
+  enabled?: boolean
+}
+export function Credits (props:CreditsProps) { }
+type LoadingProps = {
+  isLoading?: boolean,
+  hideDuration?: number,
+  labelStyle?: {},
+  showDuration?: number,
+  style?: {},
+}
+export function Loading (props: LoadingProps) { }
+
+type LegendProps = {
+  children?: Node,
+  enabled?: boolean
+}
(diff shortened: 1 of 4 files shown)


whawker:
This looks really good, just trying to figure out how to use it.

In a project that has React JSX Highcharts as a dependency, do I...

```
npm install --save-dev flow-bin

flow init
```

Then add it in the `[libs]` section of `.flowconfig`?

anajavi:
No need to modify flowconfig. Just put 
// @flow 
In the top of the file that imports the component.

whawker:
Ok cool, but the first two steps are correct?

anajavi:
Yes. The first two steps are correct.

Depending on your editor you might also add integration:
https://flow.org/en/docs/editors/

packages/react-jsx-highcharts/src/index.js.flow
```diff
@@ -0,0 +1,182 @@
...
+  children?: Node
+}
+function LegendTitle (props: LegendTitleProps) { }
+export function Legend (props: LegendProps) {
+
+}
+Legend.Title = LegendTitle;
+
+type PlotBandProps = {
+  id?: string,
```
whawker: Since v3 `id` can also be a function (returning a string), so I guess we could

```js
type Id = string | () => string

type PlotBandProps = {
  id?: Id,
  // etc
}

type Axis = {
  id?: Id,
  // etc
}
```

packages/react-jsx-highcharts/src/index.js.flow
```diff
@@ -0,0 +1,182 @@
...
+export class LineSeries extends React.Component<SeriesProps> { }
+export class PieSeries extends React.Component<SeriesProps> { }
+export class PolygonSeries extends React.Component<SeriesProps> { }
+export class PyramidSeries extends React.Component<SeriesProps> { }
+export class SankeySeries extends React.Component<SeriesProps> { }
+export class ScatterSeries extends React.Component<SeriesProps> { }
+export class SplineSeries extends React.Component<SeriesProps> { }
+export class StreamGraphSeries extends React.Component<SeriesProps> { }
+export class TreemapSeries extends React.Component<SeriesProps> { }
+export class WaterfallSeries extends React.Component<SeriesProps> { }
```
whawker: Can we also export the base type `Series`? It exported so users can use series types we don't explicitly support.

Although in the case of `Series`, prop `type` is required.

packages/react-jsx-highcharts/src/index.js.flow
```diff
@@ -0,0 +1,182 @@
...
+
+export function withHighcharts (Component: ComponentType<*>, Highcharts: HighchartsType) { }
```
whawker: Could we also export 

```js
export function provideChart (Component: ComponentType<*>) { }
export function provideAxis (Component: ComponentType<*>) { }
export function provideSeries (Component: ComponentType<*>) { }
```

packages/react-jsx-highcharts/src/index.js.flow
anajavi: I was considering the id function to be internal implementation detail, but I can add it to typings also.

packages/react-jsx-highcharts/src/index.js.flow
anajavi: I will add it.

packages/react-jsx-highcharts/src/index.js.flow
anajavi: I will add these exports too. Originally I thought them to be internal detail, but they can actually be used to make plugins for this library.

anajavi:
I will update the pull request this week.

packages/react-jsx-highcharts/src/index.js.flow
whawker: Had a looks myself but I can't figure it out. I don't get this Flow type thing at all, nevermind typing a HOC

anajavi:
There are some more problems with the typings. The functions should be declared with the "declare" keyword, otherwise flow goes into the function body to check the return type. Using the declare keyword however creates more problems elsewhere.

I'll have to let this PR sleep a little while before I can wrap my head around it.

anajavi:
I will close this for now and reopen when I have the time to look at it. Maybe flow also will get some improvements for extenal typings.

Issue #56, opened 2017-12-15
Add Highmaps support


unofang:
Thanks for your great package!

I am doing a project to visualise a big sized dataset with longitudes and latitudes. As there is no example for map-data graphs, is there anybody know how to do the basic configuration? Thank you very much !

whawker:
Unfortunately I haven't added support for Highmaps yet, sorry. It is definitely something I would like to add though!

idoromm:
Would like to add that adding map support is a must for me. This project is great but having HighMaps would make it outstanding.

whawker:
![screen shot 2018-06-15 at 07 56 43](https://user-images.githubusercontent.com/2003804/41454563-ca24868c-7071-11e8-83c2-4562e3ba03eb.png)

Making good progress on this!

whawker:
Just published the first beta, [React JSX Highmaps 0.1.0-beta.1](https://www.npmjs.com/package/react-jsx-highmaps)!

whawker:
@unofang @idoromm here is an example: https://stackblitz.com/edit/base-map-example-svixuf?file=MyMap.js

If you guys could play around with the beta and give me some feedback that would really help me out.

whawker:
React JSX Highmaps has now be released

[Release notes](https://github.com/whawker/react-jsx-highcharts/releases/tag/v3.3.0)
[NPM Package](https://www.npmjs.com/package/react-jsx-highmaps)

PR #163, opened 2018-10-13
React JSX Highmaps

whawker:
For #56. 

Just looking for any feedback, if you don't mind @anajavi ?


anajavi:
Although I'm not familiar with Highmaps, it looks very nice indeed!

I try to check it properly later, when I have the time. Now I did do only a cursory scan.

The MapLoader component left me thinking: is data loading this library's concern?
- If the data is behind authentication, the MapLoader needs to have more than url in props
- Same for POST requests and headers like Accept: application/json
- It needs fetch-polyfill to work on IE11

Would it be better to move the MapLoader to the examples? That way everybody could customise it the way they like.

Anyway, awesome work!

whawker:
Hmm yeah, fair point about the MapLoader, I think you're right in hindsight. I think I'll link to an example implementation in the readme.

I also think I'll change the default landing readme for this project, and just link out to separate packages instead.

packages/react-jsx-highmaps/.babelrc
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "env": {
+    "test": {
+      "plugins": ["@babel/transform-modules-commonjs"]
+    },
+    "production": {
+      "plugins": ["transform-react-remove-prop-types",
+        ["@babel/transform-runtime", { "useESModules": true }]
+      ]
+    }
```
anajavi: env:es is missing:
```
    "es": {
      "plugins": ["./build/use-lodash-es",
        "transform-react-remove-prop-types",
        ["@babel/transform-runtime", { "useESModules": true }]
      ]
    }
```

packages/react-jsx-highmaps/README.md
```diff
@@ -0,0 +1,100 @@
...
+We aim to pass all configuration options using the same name, so we use [Highcharts' documentation](http://api.highcharts.com/highcharts) to figure out how to achieve the same with React JSX Highcharts.
+
+### Note:
+
+There are **two** exceptions to the above;
+
+#### Exception 1
+
+Where Highcharts **events** are concerned - instead of passing `events` as an object, we use the React convention *onEventName*.
+
```
anajavi: Here an example of events would clarify what needs to be done.

packages/react-jsx-highmaps/build/use-lodash-es.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function() {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        var source = path.node.source
+        source.value = source.value.replace(/^lodash($|\/)/, 'lodash-es$1')
+      },
+    },
+  }
+}
```
anajavi: As there are no commonjs modules published in any package, we could get rid of this and lodash dependency by just using lodash-es in source code. That would simplify things.

packages/react-jsx-highmaps/src/index.js
```diff
@@ -0,0 +1,15 @@
+export * from 'react-jsx-highcharts';
+
+// Charts
+export { default as HighchartsMapChart } from './components/HighchartsMapChart';
+
+// Graph Parts
+export { default as MapNavigation } from './components/MapNavigation';
+export { default as XAxis } from './components/XAxis';
+export { default as YAxis } from './components/YAxis';
```
anajavi: XAxis and YAxis are already exported by:
```javascript
export * from 'react-jsx-highcharts';
```
Might be better to name these as MapXAxis and MapYAxis, so they don't get confused with the highcharts ones. It seems that they behave differently than the ones from react-jsx-highcharts, but I haven't tested.

packages/react-jsx-highmaps/webpack.config.js
```diff
@@ -0,0 +1,75 @@
...
+        loader: 'babel-loader',
+        exclude: /node_modules(\\|\/)(?!react-jsx-highcharts)/
+      }
+    ]
+  },
+
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env': {
+        'NODE_ENV': JSON.stringify(process.env.NODE_ENV)
```
anajavi: mode = 'production' sets this also: https://webpack.js.org/concepts/mode/

packages/react-jsx-highmaps/build/use-lodash-es.js
anajavi: It could be changed in all packages.

examples/Map/App.js
```diff
@@ -0,0 +1,69 @@
...
+
+const labelFormatter = function () {
+  if (this.point.value) {
+    return this.point.name
+  }
+}
+
+const App = () => (
+  <div className="app">
+    <Fetch url="https://code.highcharts.com/mapdata/custom/europe.geo.json">
```
anajavi: This is nice! Now there's no need to explain fetch polyfilling in the readme :)

packages/react-jsx-highmaps/src/index.js
whawker: I think I might remove `export * from 'react-jsx-highcharts';` and export only the parts that are relevant to maps. Unlike Highstock, I don't think people should be trying to build charts with this package.

packages/react-jsx-highmaps/build/use-lodash-es.js
whawker: Would this mean we could also remove the need for lodash-webpack-plugin?

packages/react-jsx-highmaps/src/components/HighchartsMapChart/HighchartsMapChart.js
```diff
@@ -0,0 +1,47 @@
...
+    const geojson = this.getGeoJSON(map);
+
+    return (
+      <BaseChart
+        chart={{ ...chart, map: geojson }}
+        mapNavigation={{ enabled: false }}
+        xAxis={{ id: 'xAxis' }}
+        yAxis={{ id: 'yAxis' }}
+        {...rest}
+        callback={this.callback(geojson)}
```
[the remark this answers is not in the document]
whawker: I believe [this modification](https://github.com/whawker/react-jsx-highcharts/pull/163/files#diff-f084ff751cb53f3465484c946894fb4bR20) should fix this problem, as `this.getGeoJSON` always returns the same object reference?

packages/react-jsx-highmaps/src/index.js
anajavi: That might be better. Could also be:
```javascript
export { withHighcharts as withHighmaps } from 'react-jsx-highcharts';
```

packages/react-jsx-highmaps/build/use-lodash-es.js
anajavi: babel-plugin-lodash could be then removed. 

lodash-webpack-plugin removes some inner parts from lodash itself, so it is still useful.

packages/react-jsx-highmaps/src/components/HighchartsMapChart/HighchartsMapChart.js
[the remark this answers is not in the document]
anajavi: Looks like it fixes the problem.

packages/react-jsx-highmaps/build/use-lodash-es.js
anajavi: I can do it, if you don't mind?

whawker: No I don't mind at all, thanks!

whawker:
Thanks very much for your help on this @anajavi, I'll publish some beta versions so we can check everything works as expect.

anajavi:
Don't you have to publish react-jsx-highcharts also as this PR had changes to it too?

whawker:
Yeah, that's the plan. Haven't published any betas for the Babel 7 stuff yet, so would be publishing those anyway.

PR #165, opened 2018-10-17
Remove height rule that breaks examples on Firefox

bondz:
The rule breaks the examples documentation page.

With the rule, the examples show like this

image

Without the rule

image

### DIFF
--- a/examples/index.css
+++ b/examples/index.css
@@ -1,5 +1,4 @@
 #root {
-  height: 100vh;
   display: flex;
   align-items: center;
   justify-content: center;


bondz:
Tested on Chrome and the rule doesn't change anything.

bondz:
The documentation needs to be rebuilt for this to work, right? @whawker

whawker:
Yes, although I'm about to publish version 3.3.0 along with a new Highmaps package, so will do it then.

bondz:
@whawker Okay, cool.

Issue #166, opened 2018-10-18
Refactor series components


anajavi:
Is there a reason why all the different Series components have their own classes, although their code is mostly the same?

Many of them could be written with a HOC like this:
```jsx
import React, { Component } from 'react';
import Series from '../Series';
import PropTypes from 'prop-types';

// This HOC returns Series component with injected type
export default function withSeriesType(seriesType, componentName) {

  class SeriesComponent extends Component {
    render () {
      return (
        <Series {...this.props} type={seriesType} />
      );
    }
  }
  SeriesComponent.propTypes = {
    id: PropTypes.oneOfType([ PropTypes.string, PropTypes.func ])
  };
  SeriesComponent.displayName = componentName;
  return SeriesComponent;
}
export const ColumnSeries = withSeriesType("column", "ColumnSeries");
export const LineSeries = withSeriesType("line", "LineSeries");
```

edit: working example

whawker:
No specific reason to be honest, probably just an oversight on my part. The HOC would definitely reduce the amount we need to test.

There are a few series types to watch out for
- [BarSeries](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/BarSeries/BarSeries.js) - Inverts the chart
- [FunnelSeries](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/FunnelSeries/FunnelSeries.js), [PyramidSeries](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/PyramidSeries/PyramidSeries.js), [PieSeries](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/PieSeries/PieSeries.js) & [VariablePieSeries](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/VariablePieSeries/VariablePieSeries.js) - Don't require a parent axis
[...]

PR #167, opened 2018-10-18
Add withSeriesType higher order component for #166

whawker:
Still need to delete the old components, and add tests. But seems to work as a proof of concept.


anajavi:
That was fast, I was just writing about the HOC :)

I will try this out as soon as I can!

I was just thinking about the HOC, what do you think about this kind of function signature:
```jsx
 export default function withSeriesType(opts = {}) {
  const { seriesType, displayName, seriesProps, propTypes } = opts;
```

whawker:
I like the idea of splitting up the props and prop types into separate arguments, however I think having to define property names  for type and display name makes it quite verbose.

i.e.
```js
const LineSeries = withSeriesType({ seriesType: 'line', displayName: 'LineSeries' })
```

What do you think of only passing one argument for type, we can lower case it for the Highchart series type, and append 'Series' on the end for the component name?

```jsx
const LineSeries = withSeriesType('AreaSplineRange')

function withSeriesType(seriesType, additionalProps = {}, additionalPropTypes = {}) {
  const SeriesComponent = props => (
    <Series {...props} {...additionalProps} type={seriesType.toLowerCase()} />
  )

  SeriesComponent.propTypes = {
    id: PropTypes.oneOfType([ PropTypes.string, PropTypes.func ]),
    ...additionalPropTypes
  };

  SeriesComponent.displayName = `${seriesType}Series`;

  return SeriesComponent;
}
```

anajavi:
You are right that there is no need for displayName. And your HOC is better, as it does not make API out of naming parameters. Let's go with that?

I must say this works as advertised. Saves 5.5kB on UMD build.

anajavi:
Testing could be done with something like this:
```jsx
import * as all from '../../../src';

Object.keys(all).filter(name => /^[A-Z].*Series$/.test(name)).forEach((SeriesName) => {
  describe(SeriesName, function ()  {
    it('renders a <Series />', function ()  {
      const wrapper = shallow(<SeriesName id="mySeries" />);
      expect(wrapper).to.have.type(Series);
    });

    it('passes other props through to <Series />', function () {
      const wrapper = shallow(<SeriesName id="myOtherSeries" data={[1, 2, 3, 4]} />);
      expect(wrapper).to.have.prop('data').eql([1, 2, 3, 4]);
    });
  });
});
```

anajavi:
Working test example:
```jsx
import React from 'react';
import * as all from '../../../src';
import Series from '../../../src/components/Series';

const skippedNames = ['BarSeries'];
Object.keys(all).filter(name => /^[A-Z].*Series$/.test(name)).forEach((seriesName) => {
  if (skippedNames.includes(seriesName)) return;

  const seriesType = seriesName.substring(0, seriesName.indexOf('Series')).toLowerCase();
  const SeriesComponent = all[seriesName];
  describe(`<${seriesName} />`, function ()  {

    it('renders a <Series />', function ()  {
      const wrapper = shallow(<SeriesComponent id="mySeries" />);
      expect(wrapper.find(Series)).to.have.type(Series);
    });
    it(`renders a <Series type="${seriesType}" />`, function () {
      const wrapper = shallow(<SeriesComponent id="mySeries" />);
      expect(wrapper).to.have.prop('type').equal(seriesType);
    });
    it('passes other props through to <Series />', function () {
      const wrapper = shallow(<SeriesComponent id="myOtherSeries" data={[1, 2, 3, 4]} />);
      expect(wrapper.find(Series)).to.have.prop('data').eql([1, 2, 3, 4]);
    });

  });
});
```

anajavi:
I moved the SeriesComponent tests to one file. Feel free to revert if you think it was a mistake.

anajavi:
Do you think there is something more to done?

whawker:
Thanks @anajavi, just pushed a commit to roll out the `withSeriesType` HOC across the other packages. I think that's everything done.

whawker:
True story! I've just published `3.3.0-beta.2` - all looks good on my end.

anajavi:
All good for me too. Thanks!

whawker:
Perfect, I'll publish the stable release versions then.

PR #169, opened 2018-10-20
Migrate from mocha to jest

anajavi:
Just testing the waters here. I find jest to be much more ergonomic than mocha/sinon/chai.

It provides out-of-the-box:
- coverage reporting
- nice mocking facilities
- expect
- snapshot testing
- fast enough to do continuous testing while writing code

This PR only takes away mocha, but keeps sinon and chai, as replacing those need some thinking.
Sinon and chai can be replaced later or kept as is if needed.

To try out coverage reporting, run:
```bash
npm run test:coverage
```
How do you feel @whawker, is this something that could be changed?


whawker:
I think migrating to Jest is a good idea. 

The current testing setup is a bit clunky, it was more a cobbling together of things that worked back when this project was only a proof of concept.

Jest is much more suited to a React project than mocha. So I'm happy.

anajavi:
So let's merge this?

Most of the changes here are made with js [jest-codemods](https://github.com/skovhus/jest-codemods), but I think they are fine.

anajavi:
The migration should be complete now.

I tried and the following works in react-jsx-highstock with version 6.2.0 (older break):
```jsx
import Highcharts from 'highcharts/highstock'
import {
  HighchartsStockChart,
  Chart,
  YAxis,
  XAxis,
  AreaSplineSeries,
  Navigator,
  withHighcharts,
  Title
} from '../../../src';
import React from 'react';
import { renderIntoDocument } from 'react-dom/test-utils'

describe('<HighchartsStockChart />', () => {
  let testContext;

  beforeEach(() => {});

  it('renders to document', (done) => {
    const addSeries = (event) => {
      expect(event.options.type).toBe('areaspline');
      done();
    }
    const Component = (props) => {
      return (<HighchartsStockChart>
        <Chart zoomType="x" onAddSeries={addSeries}/>

        <Title>Highstocks Example</Title>

        <XAxis>
          <XAxis.Title>Time</XAxis.Title>
        </XAxis>

        <YAxis>
          <YAxis.Title>Price</YAxis.Title>
          <AreaSplineSeries id="profit" name="Profit" data={[1,2,3,4]}/>
        </YAxis>

        <Navigator>
          <Navigator.Series seriesId="profit"/>
          <Navigator.Series seriesId="twitter"/>
        </Navigator>
      </HighchartsStockChart>);
    };
    const WithComponent = withHighcharts(Component, Highcharts);
    const renderedChart = renderIntoDocument(<WithComponent />);
  });
});
```

So it should be possible to trigger regressions and changes in real highcharts code.

whawker:
Wow, excellent work on this! 😮💯 

I agree we should add tests similar to your previous comment, as they would provide a level of integration tests that we currently don't have.

anajavi:
Also, enzyme seems to be lagging behind on support for react 16, so different type of testing might be needed for some components (Contexts for example).

PR #175, opened 2018-10-31
WIP: highcharts-gantt

jvskriubakken:
Setting up some structure based on highmaps


packages/react-jsx-highcharts-gantt/src/index.js
```diff
@@ -0,0 +1 @@
+// TODO: Is it here that the "common modules should be exported (Title, Tooltip, etc)" ?
```
whawker: Yep

packages/react-jsx-highcharts-gantt/src/index.js
```diff
@@ -0,0 +1,24 @@
...
+} from 'react-jsx-highcharts';
+
+
+// Charts
+export { default as HighchartsGanttChart } from './components/HighchartsGanttChart';
+
+// Graph Parts
+
+export { default as XAxis } from './components/XAxis'; // TODO: ?
+export { default as YAxis } from './components/YAxis'; // TODO: ?
```
whawker: These will be exactly the same as the XAxis and YAxis in `react-jsx-highmaps` (we can clean that up at a later date)

packages/react-jsx-highcharts-gantt/src/components/Pathfinder/Pathfinder.js
```diff
@@ -0,0 +1,29 @@
...
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      rendered: false
+    };
+  }
+
+  render () {
+    // TODO: Howto?
```
whawker: [This component](https://github.com/whawker/react-jsx-highcharts/blob/d3257a8ed0f30581ed90210b69fb8071596e629d/packages/react-jsx-highcharts/src/components/Legend/Legend.js) might help.

You need to [provide the chart](https://github.com/whawker/react-jsx-highcharts/blob/d3257a8ed0f30581ed90210b69fb8071596e629d/packages/react-jsx-highcharts/src/components/Legend/index.js#L3-L4) to the component

Then `chart.update` with new `pathfinder` config on [`componentDidMount`](https://github.com/whawker/react-jsx-highcharts/blob/d3257a8ed0f30581ed90210b69fb8071596e629d/packages/react-jsx-highcharts/src/components/Legend/Legend.js#L20-L22) and [`componentDidUpdate`](https://github.com/whawker/react-jsx-highcharts/blob/d3257a8ed0f30581ed90210b69fb8071596e629d/packages/react-jsx-highcharts/src/components/Legend/Legend.js#L26-L28). 

Finally, [disable](https://github.com/whawker/react-jsx-highcharts/blob/d3257a8ed0f30581ed90210b69fb8071596e629d/packages/react-jsx-highcharts/src/components/Legend/Legend.js#L33) the pathfinder on unmount.

Then component will have no children, so  you can just `return null` from render

packages/react-jsx-highcharts-gantt/src/components/Pathfinder/index.js
```diff
@@ -0,0 +1,4 @@
+import Pathfinder from './Pathfinder';
+export default Pathfinder;
+
+// TODO: Does Pathfinder need sub components for the types of the properties: startMarker, endMarker and marker?
```
whawker: > The component will have no children, so you can just return null from render

Actually ignore what I said above, probably does make sense to make sub components out of these.

I that case, the above Pathfinder component should  `return this.props.children` instead

jvskriubakken: You mean that I should only ignore what you said about "the component will have no children"?

whawker: Yeah, sorry that could have been clearer!

jvskriubakken: No worries, just asked to be sure... trying to implement your other comments now. Thanks for your great help so far!

whawker: No problem, great to see it coming along!

jvskriubakken:
Looking at the Pathfinder API (https://api.highcharts.com/gantt/pathfinder)
I'm not sure whether I must expose:
`ChartPathfinder.Marker`
in Pathfinder/index.js
or
`
ChartPathfinder.startMarker = provideChart(Marker);
ChartPathfinder.marker = provideChart(Marker);
ChartPathfinder.endMarker = provideChart(Marker);
`

Also I'm wondering if there is a way for me to actually easy test the result of what I'm making visually? Like the examples I've started on?

whawker:
> I'm not sure whether I must expose:
ChartPathfinder.Marker
in Pathfinder/index.js
or
ChartPathfinder.startMarker = provideChart(Marker); ChartPathfinder.marker = provideChart(Marker); ChartPathfinder.endMarker = provideChart(Marker);

I would suggest making a Pathfinder.Marker component that accepts a `markerType` prop that defines which marker to update

The component below would update [`pathfinder.marker`](https://api.highcharts.com/gantt/pathfinder.marker) because of the default prop `markerType = 'marker'`
```jsx
import React, { Component } from 'react';
import PropTypes from 'prop-types';
import { attempt } from 'lodash-es';
import { getModifiedProps } from 'react-jsx-highcharts';

class PathfinderMarker extends Component {

  static propTypes = {
    getChart: PropTypes.func.isRequired, // Provided by ChartProvider
    enabled: PropTypes.bool.isRequired
  };

  static defaultProps = {
    enabled: true
    markerType: 'marker'
  };

  componentDidMount () {
    const { children, ...rest } = this.props;
    this.updatePathfinderMarker({
      ...rest
    });
  }

  componentDidUpdate (prevProps) {
    const modifiedProps = getModifiedProps(prevProps, this.props);
      this.updatePathfinderMarker(modifiedProps);
    }
  }

  componentWillUnmount () {
    attempt(this.updatePathfinderMarker, { enabled: false });
  }

  updatePathfinderMarker = config => {
    const { getChart, markerType } = this.props
    const chart = getChart();
    chart.update({
      pathfinder: {
        [markerType]: config
      }
    }, true);
  }

  render () {
    return null
  }
}

export default PathfinderMarker;
```

Then you can make a `PathfinderStartMarker` and `PathfinderEndMarker` component, that wraps `PathFinderMarker` and overrides the default `markerType`

```jsx
const PathfinderStartMarker = ({ ...props }) => (
  <PathfinderMarker symbol='diamond' {...props} markerType='startMarker' />
)
export  default PathfinderStartMarker
```

Then you `Pathfinder/index.js` becomes

```js
import Pathfinder from './Pathfinder';
import PathfinderMarker from './PathfinderMarker';
import PathfinderStartMarker from './PathfinderStartMarker';
import PathfinderEndMarker from './PathfinderEndMarker';
import { provideChart } from 'react-jsx-highcharts'
const ChartPathfinder = provideChart(Pathfinder);
Pathfinder.Marker = provideChart(PathfinderMarker);
Pathfinder.StartMarker = provideChart(PathfinderStartMarker);
Pathfinder.EndMarker = provideChart(PathfinderEndMarker);
export default ChartPathfinder;
```

packages/react-jsx-highcharts-gantt/src/components/Pathfinder/Pathfinder.js
```diff
@@ -0,0 +1,57 @@
...
+  static defaultProps = {
+    enabled: true
+  };
+
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      rendered: false
+    };
```
whawker: This state isn't used anywhere, so this constructor is unnecessary

packages/react-jsx-highcharts-gantt/src/components/Pathfinder/Pathfinder.js
```diff
@@ -0,0 +1,57 @@
...
+
+  static propTypes = {
+    getChart: PropTypes.func.isRequired, // Provided by ChartProvider
+    enabled: PropTypes.bool.isRequired,
+    dashStyle: PropTypes.string(),
+    lineWidth: PropTypes.number()
+  };
+
+  static defaultProps = {
+    enabled: true
```
whawker: These default props should mirror the [pathfinder api](https://api.highcharts.com/gantt/pathfinder.dashStyle) defaults

I.e.

```js
static defaultProps = {
  enabled: true,
  dashStyle: 'solid',
  lineWidth:  1,
  type: 'straight|simpleConnect'
}
```

whawker:
> Also I'm wondering if there is a way for me to actually easy test the result of what I'm making visually? Like the examples I've started on?

This is a little bit more complicated, might be best if I set this up for you. I'll make a PR in your repo

jvskriubakken:
> > Also I'm wondering if there is a way for me to actually easy test the result of what I'm making visually? Like the examples I've started on?
> 
> This is a little bit more complicated, might be best if I set this up for you. I'll make a PR in your repo

Great! Thanks!

packages/react-jsx-highcharts-gantt/src/components/Pathfinder/Pathfinder.js
jvskriubakken: Sure, copy and paste.

jvskriubakken:
According to the API doc it seems to be a small distinction between the property marker and the properties startMarker and endMarker. startMarker and endMarker has the property symbol, but the marker does not. Would this become a problem with our current design?

whawker:
I don't think it should no, I think Highcharts will just ignore symbol for `marker` config.

jvskriubakken:
I've merged your pull request with examples.

I've sucessfully done:
```
cd packages/react-jsx-highcharts-gantt
npm run build:prod
```
But 
```
cd ../../examples
npm run build:examples
```
fails with 
```
> webpack --config webpack.config.js

sh: webpack: command not found
```
I've tried to install both webpack and webpack-cli globally,
but then it fails with
```
> webpack --config webpack.config.js

/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/webpack-cli/bin/cli.js:244
                                throw err;
                                ^

Error: Cannot find module 'webpack'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Object.<anonymous> (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/webpack-cli/bin/convert-argv.js:7:24)
    at Module._compile (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at yargs.parse (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/webpack-cli/bin/cli.js:241:14)
    at Object.parse (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/yargs/yargs.js:563:18)
    at /Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/webpack-cli/bin/cli.js:219:8
    at Object.<anonymous> (/Users/jorund/IdeaProjects/react-jsx-highcharts-fork/examples/node_modules/webpack-cli/bin/cli.js:538:3)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at module.exports.filename (/usr/local/lib/node_modules/webpack-cli/node_modules/import-local/index.js:16:66)
    at /usr/local/lib/node_modules/webpack-cli/bin/cli.js:13:6
    at Object.<anonymous> (/usr/local/lib/node_modules/webpack-cli/bin/cli.js:538:3)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)

```

whawker:
Apologies, you will need to run `npm install` in the examples directory.

Looking over the currently examples code, they will likely not work, I think you need to
1. Export `withHighcharts` from [here](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/packages/react-jsx-highcharts-gantt/src/index.js#L9)
1. Import it in [here](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/HighchartsGanttSimple/App.js#L3) (as currently it is undefined)
 1. Also you'll need to `import Highcharts from 'highcharts/highcharts-gantt` [here](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/HighchartsGanttSimple/App.js#L2)

jvskriubakken:
Posting recipe for building examples for later use.

To build the examples, firstly you'll need to compile the library you're working on

```bash
cd packages/react-jsx-highcharts-gantt
npm run build:prod
```

Then you need to navigate to the `examples` directory, compile them, add serve them locally
```bash
# Assuming you're in the packages/react-jsx-highcharts-gantt directory
cd ../../examples
npm run build:examples
cd ..
# npm install -g http-server if you don't have it installed
http-server 
# Navigate to http://localhost:8080/examples/<Example Name>/index.html
```

Where currently, `<Example Name>` would be either `HighchartsGanttSimple` or `HighchartsGanttWithSubtasks`

jvskriubakken:
I've successfully built highcharts-gantt and examples and started the http-server.

I can access other examples graphs, but the grantt graphs fails with the error:
```
Uncaught ReferenceError: ReactHighcharts is not defined
    at Object.<anonymous> (bundle.js:1)
    at n (bundle.js:1)
    at Module.<anonymous> (bundle.js:1)
    at n (bundle.js:1)
    at bundle.js:1
    at bundle.js:1
```
Maybe you could help med with this plumbing problem?

whawker:
That's odd, I've just cloned your repo and following the recipe above I see the following.

screen shot 2018-11-02 at 08 33 53

The other examples are  likely working because they're [loading their library from unpkg.com](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/webpack.config.js#L71-L73)

Are you definitely running `http-server` from the root directory? i.e. one directory above the `examples` directory. Otherwise [the relative path](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/webpack.config.js#L70) won't work correctly

```sh
cd ../../examples
npm run build:examples
cd .. # <-- this step
http-server 
```

jvskriubakken:
Thanks! You where correct, I did not to the cd .. :(

jvskriubakken:
I got the same result as you now, after starting the http server from the root directory.

But now I am lost, as the chart renders blank and now errors. Am I setting up the chart correctly in my examples? Or could is something missing in the implementation of HighchartsGanttChart or GanttSeries?

jvskriubakken:
Could it be a missing dependency on highcharts-gantt.js ?

whawker:
I think all you need to do is provide the [axis context to your series](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highmaps/src/index.js#L29-L30)?

jvskriubakken:
OK. When I connect the yAxis as the parentAxis, then it fails 
```
const parentAxisId = { axisId: 'yAxis' };
export const GanttSeries = withSeriesType('gantt', parentAxisId);
```
Then it fails with highchart errors 11 and 19:
https://www.highcharts.com/errors/11
https://www.highcharts.com/errors/19

```
Error: Highcharts error #11: www.highcharts.com/errors/11
    at Object.a.error (highcharts-gantt.src.js:426)
    at F.setTickInterval (highcharts-gantt.src.js:17748)
    at F.setTickInterval (highcharts-gantt.src.js:45694)
    at F.a.(anonymous function) [as setTickInterval] (https://code.highcharts.com/gantt/6.2.0/highcharts-gantt.js:20:186)
    at F.setScale (highcharts-gantt.src.js:18365)
    at highcharts-gantt.src.js:26256
    at Array.forEach (<anonymous>)
    at a.each (highcharts-gantt.src.js:2285)
    at a.Chart.redraw (highcharts-gantt.src.js:26250)
    at a.Chart.<anonymous> (highcharts-gantt.src.js:34994)
```

jvskriubakken:
Is it correct to force yAxis as parentAxis to GanttSeries?

whawker:
This is where we get to the point where we need to read the Highcharts source code and figure out the nuances.

### XAxis
#### Changes:
1. Enforced type is [datetime](https://github.com/highcharts/highcharts/blob/f10463e11d153df7effd726cde972822dac8b771/js/parts-gantt/GanttChart.js#L55-L57)
1. Grid is [enabled by default](https://github.com/highcharts/highcharts/blob/f10463e11d153df7effd726cde972822dac8b771/js/parts-gantt/GanttChart.js#L48-L52), and opposite is true (can be overriden)
```jsx
import { XAxis } from 'react-jsx-highcharts';

const GanttXAxis = ({ ...props }) => (
  <XAxis {...props} id="xAxis" dynamicAxis={false} type='datetime' />
);

GanttXAxis.defaultProps = {
  grid: { enabled: true },
  opposite: true,
  visible: false
};

export default GanttXAxis;
```

### YAxis
#### Changes:
1. [Added Default options](https://github.com/highcharts/highcharts/blob/f10463e11d153df7effd726cde972822dac8b771/js/parts-gantt/GanttChart.js#L66-L72)
2. [Type is `treegrid` unless there are categories](https://github.com/highcharts/highcharts/blob/f10463e11d153df7effd726cde972822dac8b771/js/parts-gantt/GanttChart.js#L76)
```jsx
import React from 'react';
import { YAxis } from 'react-jsx-highcharts';

const GanttYAxis = ({ categories, type, ...props }) => {
  const seriesType = categories ? type : 'treegrid';
  return (
    <YAxis {...props} id="yAxis" dynamicAxis={false} type={seriesType} />
  );
}

GanttYAxis.defaultProps = {
  grid: { enabled: true },
  reversed: true,
  staticScale: 50,
  visible: false
};

export default GanttYAxis;
```

jvskriubakken:
Great! We have breakthrough :-)

whawker:
It would be good to add an example using XRange series and pathfinder

> `highcharts-gantt.src.js:5314 Error: <rect> attribute y: Expected length, "NaN".`

Not sure about this one, will have a  look into when this is merged and I get time.

packages/react-jsx-highcharts-gantt/src/components/YAxis/YAxis.js
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <YAxis {...props} id="yAxis" dynamicAxis={false} type={seriesType} />
+  );
+}
+
+GanttYAxis.defaultProps = {
+  grid: { enabled: true },
+  reversed: true,
+  staticScale: 50,
+  visible: false
```
whawker: Apologies I made a mistake here, can you remove the visible default prop from both axes. 

They should both be [visible true](https://api.highcharts.com/gantt/yAxis.visible) (which is already default)

jvskriubakken:
> It would be good to add an example using XRange series and pathfinder
> 
> > `highcharts-gantt.src.js:5314 Error: <rect> attribute y: Expected length, "NaN".`
> 
> Not sure about this one, will have a look into when this is merged and I get time.

OK. Thanks for helping me with this. I will add more examples, covering both Pathfinder and XRange. Have a great weekend!

jvskriubakken:
Hi @whawker 

I've added several more examples:
HighchartsGanttInvertedChart
Fails with a react error 31

HighchartsGanttWithHighstockNavigation
Fails with error:  Cannot read property 'Series' of undefined

HighchartsGanttProjectManagement
Renders fine.

HighchartsGanttTweakASingleDependency
This example uses the Pathfinder component.
Fails with a react error 152

I have yet not found any example using the XRange series.

jvskriubakken:
Added another example: HighchartsGanttResourceManagement
which fails with same error as HighchartsGanttWithHighstockNavigation:
"series is not defined"

I think I need some help to be able to come past this problems. Do you have the time @whawker ?

examples/HighchartsGanttInvertedChart/App.js
```diff
@@ -0,0 +1,86 @@
...
+            {
+              [
+                {
+                  opposite: false
+                },
+                {
+                  opposite: false
+                }
+              ]
+            }
```
whawker: This isn't how React JSX Highcharts works. This should be `<XAxis opposite={false}>`.

Plus we've only allowed for a single XAxis at the moment, so we need to revisit the XAxis component.

examples/HighchartsGanttWithHighstockNavigation/App.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+    return (
+      <div className="app">
+        <HighchartsGanttChart>
+          <Title>Gantt Chart with Highstock Navigation</Title>
+          <YAxis
+            uniqueNames={true}
+          />
+          <Navigator enabled={true}>
```
whawker: As stated in the readme, [`enabled=true`](https://github.com/whawker/react-jsx-highcharts#example) is assumed on mount. So this is redundant

examples/HighchartsGanttWithHighstockNavigation/App.js
```diff
@@ -0,0 +1,88 @@
...
+
+  render() {
+
+
+    return (
+      <div className="app">
+        <HighchartsGanttChart>
+          <Title>Gantt Chart with Highstock Navigation</Title>
+          <YAxis
+            uniqueNames={true}
```
whawker: True is the default argument for a React prop, so this can just be `<YAxis uniqueNames>`

whawker:
> HighchartsGanttWithHighstockNavigation
Fails with error: Cannot read property 'Series' of undefined

This is probably because both libraries are exposed on the [same `React Highcharts` global](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/webpack.config.js#L136-L139), This is probably because we've never used two different modules before. I can look into this

examples/HighchartsGanttInvertedChart/App.js
jvskriubakken: OK. Thank for the input. I felt I was doing something wrong. It was this original Higcharts code that confused me:
```
xAxis: [{
        opposite: false
    }, {
        opposite: false
}]
```

examples/HighchartsGanttWithHighstockNavigation/App.js
jvskriubakken: OK, I've removed these.

examples/HighchartsGanttWithHighstockNavigation/App.js
jvskriubakken: OK, I've removed "={true}" here and other similar places, so that the code is easier to read.

jvskriubakken:
> > HighchartsGanttWithHighstockNavigation
> > Fails with error: Cannot read property 'Series' of undefined
> 
> This is probably because both libraries are exposed on the [same `React Highcharts` global](https://github.com/jvskriubakken/react-jsx-highcharts/blob/highcharts-gantt/examples/webpack.config.js#L136-L139), This is probably because we've never used two different modules before. I can look into this

Great! Thanks!

whawker:
> HighchartsGanttResourceManagement
Renders partly OK, but table on left is rendered incorrectly (YAxis with type "category"). Correct rendering can be seen here:
https://www.highcharts.com/gantt/demo/resource-management

Just looking into this, I think we might need to introduce new `Grid` and `Column` components that create additional axes.

jvskriubakken:
> > HighchartsGanttResourceManagement
> > Renders partly OK, but table on left is rendered incorrectly (YAxis with type "category"). Correct rendering can be seen here:
> > https://www.highcharts.com/gantt/demo/resource-management
> 
> Just looking into this, I think we might need to introduce new `Grid` and `Column` components that create additional axes.

Like a new YAxis spescially for gantt? That supports a new component Column as children?

whawker:
The YAxis should be able to stay as it is. What I mean is, instead of 

```jsx
<YAxis
  type="category"
  grid={
    {
      enabled: true,
      columns: [
        {
          title: {
            text: 'Model'
          },
          categories: map(series, function (s) {
            return s.name;
          })
        },
        {
          title: {
            text: 'Rented To'
          },
          categories: map(series, function (s) {
            return s.current.rentedTo;
          })
        },
        {
          title: {
            text: 'From'
          },
          categories: map(series, function (s) {
            return dateFormat('%e. %b', s.current.from);
          })
        },
        {
          title: {
            text: 'To'
          },
          categories: map(series, function (s) {
            return dateFormat('%e. %b', s.current.to);
          })
        }
      ]
    }
  }
/>
```

We would have

```jsx
<YAxis type="category">
  <Grid>
    <Column categories={series.map(s => s.name)}>
      <Column.Title>Model</Column.Title>
    </Column>
    <Column categories={series.map(s => s.current.rentedTo)}>
      <Column.Title>Rented To</Column.Title>
    </Column>
    <Column categories={series.map(s => dateFormat('%e. %b', s.current.from))}>
      <Column.Title>From</Column.Title>
    </Column>
    <Column categories={series.map(s => dateFormat('%e. %b', s.current.to))}>
      <Column.Title>To</Column.Title>
    </Column>
  </Grid>
</YAxis>
```

I looks as if this new `Column` component would be very similar to an Axis component (even using `chart.addAxis` under the hood).

I can implement if you like?

jvskriubakken:
Yes, please go a head and implement!

whawker:
This is proving to be very difficult, seems Highcharts Gantt charts do not want to place nicely with dynamically added axes.

The "Resource management" use case might have to be something we can't support...

jvskriubakken:
> This is proving to be very difficult, seems Highcharts Gantt charts do not want to place nicely with dynamically added axes.
> 
> The "Resource management" use case might have to be something we can't support...

That was unfortunate. Maybe we could try make Higcharts support that in the future. As a middleway, would it be possible to just have a Grid-component which takes an array of columns?

jvskriubakken:
My personal need, is to get the following examples working:
HighchartsGanttWithHighstockNavigation and HighchartsGanttCategorisedYAxisWithGrouping. But I also this one would be nice to have working: HighchartsGanttTweakASingleDependency (for the sake of the gantt-support in highcharts-jsx)

jvskriubakken:
> Update,
> 
> I have * HighchartsGanttResourceManagement* working with proper React components
> screen shot 2018-11-19 at 09 20 16
> 
> Will look tonight at the _HighchartsGanttWithHighstockNavigation_ issues.

That is fantastic Will!

whawker:
This is not complete, but merging to a `react-jsx-highcharts-gantt` branch so I can continue working on it.

PR #176, opened 2018-11-01
Redraw only once on mount, remove forced updates

anajavi:
This saves a couple of unnecessary redraw calls when adding axis or series. I found the calls while doing profiling.

It does it by chart.addAxis and chart.addSeries calls to have redraw = false. Then it relies on the last call to addEventProps to call updateFn with redraw = true.

Also this treats render() as lifecycle method so that Axis and Series don't need to call this.forceUpdate.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Axis/Axis.js
+++ b/packages/react-jsx-highcharts/src/components/Axis/Axis.js
@@ -60,27 +60,29 @@ class Axis extends Component {
     }
   }
 
-  render () {
-    if (!this.axis) {
-      const { id, dynamicAxis, isX, getChart } = this.props;
-      const chart = getChart();
+  createAxis () {
+    const { id, dynamicAxis, isX, getChart } = this.props;
+    const chart = getChart();
 
-      // Create Highcharts Axis
-      const opts = this.getAxisConfig();
-      if (dynamicAxis) {
-        this.axis = chart.addAxis(opts, isX, false);
-      } else {
-        // ZAxis cannot be added dynamically, Maps only have a single axes - update instead
-        const axisId = isFunction(id) ? id() : id
-        this.axis = chart.get(axisId);
-        this.axis.update(opts, false);
-      }
-
-      const update = this.axis.update.bind(this.axis);
-
-      // we rely addEventProps to call redraw
-      addEventProps(update, this.props, true);
+    // Create Highcharts Axis
+    const opts = this.getAxisConfig();
+    if (dynamicAxis) {
+      this.axis = chart.addAxis(opts, isX, false);
+    } else {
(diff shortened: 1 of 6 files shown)


packages/react-jsx-highcharts/src/components/Axis/Axis.js
```diff
@@ -83,7 +61,26 @@ class Axis extends Component {
   }
 
   render () {
-    if (!this.axis) return null;
+    if (!this.axis) {
+      const { id, dynamicAxis, isX, getChart } = this.props;
```
whawker: Interesting, good spot on this! But having this logic in render makes me feel uneasy.

I know it is exactly the same meaning but could we extract this out into...

```jsx
render () {
  if (!this.axis) this.initialRender() // line 65-83 in this function
  // As before
}

Comment (anajavi) [reply packages/react-jsx-highcharts/src/components/Axis/Axis.js]: I agree that setting instance variables in render() seems like an anti-pattern. Unfortunately, there is no better lifecycle method to do that.

I extracted methods: createAxis() and createSeries(). Those names felt more explicit.

File: packages/react-jsx-highcharts/src/components/Axis/Axis.js [hunk 2]
```diff
@@ -82,8 +60,29 @@ class Axis extends Component {
     }
   }
 
+  createAxis () {
```
whawker: Think this method should be bound to `this`

```suggestion
  createAxis = () => {
```

packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -102,8 +88,21 @@ class Series extends Component {
     return config;
   }
 
+  createSeries () {
```
whawker: Ditto with this method, it should be bound to `this`

```suggestion
  createSeries = () => {
```

anajavi:
Ultimately, it would be nice to call chart.redraw() on initial mount only when all series children have been added, but I can't come up with a way to do that cleanly.

whawker:
Potentially that could be done with an [`addSeries`](https://api.highcharts.com/highcharts/chart.events.addSeries) event that we debounce, so only the final addSeries event triggers a redraw?

packages/react-jsx-highcharts/src/components/Axis/Axis.js
anajavi: Why is that? It is not passed as parameter to anywhere.

anajavi:
But how do we know which addSeries event is final?

whawker:
Well we don't need to know which series is last, all we care about is that the `addSeries` handler is no  longer being triggered.

[Example here](https://stackblitz.com/edit/base-chart-example-e9uqbc?file=index.js)

PR #179, opened 2018-11-03
Batch Axis and Series redraws

### LINKED ISSUE #177: Redraw chart only after all series are added
#### Bug or Feature Request?
Performance improvement.

#### Description
With lots of series and datapoints in them, mounting chart can be slow. This is because every series does redraw when it mounts.

> Potentially that could be done with an [`addSeries`](https://api.highcharts.com/highcharts/chart.events.addSeries) event that we debounce, so only the final addSeries event triggers a redraw?

> Well we don't need to know which series is last, all we care about is that the `addSeries` handler is no longer being triggered.
> 
> [Example here](https://stackblitz.com/edit/base-chart-example-e9uqbc?file=index.js)

#### Versions
* React JSX Highcharts version: 3.3.0
* Highcharts version: 6.2.0
* React version: 16.6

anajavi:
This makes Axis and Series mounting and updating to do less redraws.

Fixes #177.

Tests are still failing. Will do those if this is otherwise suitable. Naming of needsRedraw is one thing I am not sure about.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -75,7 +75,8 @@ class Series extends Component {
   componentWillUnmount () {
     if (this.series && this.series.remove) {
       // Series may have already been removed, i.e. when Axis unmounted
-      attempt(this.series.remove.bind(this.series));
+      attempt(this.series.remove.bind(this.series), false);
+      this.props.needsRedraw();
     }
   }
 
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -62,6 +62,7 @@ class Series extends Component {
     const modifiedProps = getModifiedProps(prevProps, rest);
     if (modifiedProps !== false) {
       this.series.update(modifiedProps, false);
+      needsRedraw = true;
     }
     if (needsRedraw) {
       this.props.needsRedraw();
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -19,7 +19,8 @@ class Series extends Component {
     data: PropTypes.any,
     visible: PropTypes.bool,
     getChart: PropTypes.func, // Provided by ChartProvider
-    getAxis: PropTypes.func, // Provided by AxisProvider
+    needsRedraw: PropTypes.func, // Provided by ChartProvider
+    getAxis: PropTypes.func // Provided by AxisProvider
   };
 
   static defaultProps = {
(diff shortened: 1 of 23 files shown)


packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -44,20 +45,30 @@ class Series extends Component {
...
     if (visible !== prevProps.visible) {
-      this.series.setVisible(visible);
+      this.series.setVisible(visible, false);
+      needsRedraw = true;
     }
 
     const modifiedProps = getModifiedProps(prevProps, rest);
     if (modifiedProps !== false) {
-      this.series.update(modifiedProps);
+      this.series.update(modifiedProps, false);
```
whawker: Think we need a `needsRedraw = true` after this line?

anajavi: Oops..

anajavi:
I believe this PR is now finalized.

Now it is down to 5 redraws from 96. Full render went from 3.5sec to under a second. I am very happy with the results.

I would have added batching tests in basechart, but I was stopped by this issue in jest timers https://github.com/facebook/jest/issues/3465. I believe this will be fixed in Jest 24, due this december. 

@whawker do you think this can now be merged and cut into an alpha release?

whawker:
Apologies, been travelling a lot these past few days. Not had much chance to sit down with my laptop. Looking over this now.

whawker:
Looks good to me, nice work. I'll push out an alpha release.

I feel this should be a patch release, rather than an minor?

anajavi:
No worries, you are very fast by normal pull request standards :)

Patch as there are no new features. Minor as it changes things quite a bit.

Maybe a minor version, because it has a chance to break something? For example somebody might have onRender events bound, now they fire only once.

whawker:
I've published `3.3.1-alpha.1` (only react-jsx-highcharts at this point) however there seems to be an issue with the [Sankey example](https://stackblitz.com/edit/github-4n2yqq?file=src/index.js)? All other examples seem fine though

Looks as though Highcharts is removing the series when redraw is false?

anajavi:
Looks like it could be [this line](https://github.com/highcharts/highcharts/blob/master/js/modules/sankey.src.js#L607).

I will check it later.

anajavi:
Reproduction with bare highcharts: https://jsfiddle.net/cm352rLt/7/

Issue sent to highcharts: https://github.com/highcharts/highcharts/issues/9300

The problem seems to be that when series is updated, it is removed first:
https://github.com/highcharts/highcharts/blob/ee6f7a4b5ce85eb3c6145d66c31d32a631f07666/js/parts/Dynamics.js#L1096

And sankey series has no "isInitialized" check in destroy().

I can write some kind of bailout for sankey series.

Also it might be good to put the event handler update to addSeries call, should be faster.

anajavi:
Are you okay with the hack I made on https://github.com/whawker/react-jsx-highcharts/commit/c0d0ee8a15a351538a2b5862e7b9bc3fada70655

It is quite ugly.

If it is too ugly, I can add batchRedraw prop to Series, and add it via withSeriesType HOC.

With any luck this is resolved in Highcharts 7 and the hack can be removed..

anajavi:
Sankey is already fixed in highcharts master by https://github.com/highcharts/highcharts/commit/593f100895141e33aa4fff3fb903acdf61eb2de0

PR #184, opened 2018-11-08
Add eventhandlers to Axis and Series when they are created

anajavi:
It's pretty straight forward, but I don't know the history of adding the event handlers with update, so decided to ask for review.

And this is not urgent by any means.

edit: it also seems to fix sankey series problem, removed the hack for it

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Axis/Axis.js
+++ b/packages/react-jsx-highcharts/src/components/Axis/Axis.js
@@ -4,7 +4,7 @@ import uuid from 'uuid/v4';
 import { isFunction } from 'lodash-es';
 import { attempt } from 'lodash-es';
 import { Provider } from '../AxisContext';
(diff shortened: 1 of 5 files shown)


anajavi:
I've thought that there is a good reason to add event handlers to chart with addEventHandlersManually. If there isn't, it's quite simple to fix?

anajavi:
Changing Chart to do:
```javascript
  componentDidMount () {
    const { getHighcharts, getChart, needsRedraw, children, ...rest } = this.props;
    const notEventProps = getNonEventHandlerProps(rest);

    this.updateChart({
      ...notEventProps,
      events: getEventsConfig(rest)
    });
  }
```
... hangs the new seriestype tests, which rely on Chart events. So there is a reason to do them one by one.

I will merge this now. The Chart needs more investigation.

Issue #171, opened 2018-10-22
Planning for next major release


anajavi:
Highcharts 7 should be here soon, so I thought the next major release could be planned.

Here's my list of breaking changes I like to see for discussion:
1. Drop equality checks from prop changes, do shallow check instead. This should give better performance
2. Drop support for Highcharts 5 (what about dropping 6 too?)
3. Add support for Highcharts 7
4. Drop support for React < 16.3. Enables the use of new lifecycle methods, if needed
5. Remove FlagSeries as it is now called FlagsSeries

whawker:
Good idea planning for the next release.

I'm happy with those suggestions, except for dropping React < 16.3 support. 

The main use I have for this project is at my place of work, where it is part of large React web app currently using React 16.2. We do plan to update to the latest version of the React, but I can't see us doing that before year end, due to other commitments. It looks as if Highcharts 7 will come out mid November.

Other features I'd like to consider are
* Proper [`colorAxis`](https://api.highcharts.com/highcharts/colorAxis) support.
* Improved responsive capabilities

anajavi:
What does improved responsive capabilities mean?

whawker:
There is a [responsive example](https://whawker.github.io/react-jsx-highcharts/examples/Responsive/index.html), but to be honest it is pretty poor. Trying to do anything more complicated doesn't particularly work very well.

I wonder if there is something we can do with [`responsive.rules`](https://api.highcharts.com/highcharts/responsive) instead.

anajavi:
Wrapping the whole chart in something like https://github.com/souporserious/react-measure might do it.
That could be wired to do setSize on chart, and making the responsive.rules to run.

anajavi:
I ran a quick test with Highcharts 7 from the repository. Nothing seems to break yet.

whawker:
Great news.

I think with the number of annotations Highstock 7 adds, we should add better support for them.

anajavi:
Something like this?
```jsx
import React, { Component, Children, cloneElement, isValidElement } from 'react';
import PropTypes from 'prop-types';
import uuid from 'uuid/v4';
import { isFunction } from 'lodash-es';
import { attempt } from 'lodash-es';
import Hidden from '../Hidden';
import getModifiedProps from '../../utils/getModifiedProps';

class Annotation extends Component {

  static propTypes = {
    getChart: PropTypes.func.isRequired // Provided by ChartProvider
  };

  static defaultProps = {
[...]

PR #189, opened 2018-11-12
WIP: Next major version

anajavi:
**Breaking changes:**

- [X] Drop immutable, lodash.isEqual
- [ ] Make Series & friends PureComponent
- [ ] Drop Highcharts < 7
- [ ] Drop React < 16.2
- [ ] Remove FlagSeries as it is now called FlagsSeries
- [ ] Drop support for IE < 11
  - [ ] babel-preset-env target IE = 11
- [ ] replace lodash with native functions where convenient
  - [X] forEach
  - [X] isFunction
  - [ ] pickBy
  - [ ] omitBy
  - [X] omit (to be removed from [lodash](https://github.com/lodash/lodash/wiki/Roadmap) )
  - [ ] attempt
  - [ ] find (likely removed from [lodash](https://github.com/lodash/lodash/commit/4ba0f8b6124062836af257dae8d208976b0e05b7#comments)
  - [ ] findKey
  - [ ] defaultTo
  - [ ] lowerFirst
  - [ ] mapKeys
  - [X] size
  - [ ] wrap (to be removed from [lodash](https://github.com/lodash/lodash/wiki/Roadmap) )
  - [ ] isPlainObject
  - [ ] isEmpty
  - [ ] findIndex
  - [ ] upperFirst

**Features:**
- [ ] Annotation component
- [ ] colorAxis support
- [ ] better responsiveness (yet to be defined)

**Other:**
- [ ] Update examples
- [ ] Update docs

As discussed on #171

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -1,17 +1,14 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import uuid from 'uuid/v4';
 import { isFunction } from 'lodash-es';
-import { isEqual } from 'lodash-es';
 import { attempt } from 'lodash-es';
-import isImmutable from 'is-immutable';
-import immutableEqual from 'immutable-is';
 import { Provider } from '../SeriesContext';
 import { getNonEventHandlerProps, getEventsConfig } from '../../utils/events';
 import getModifiedProps from '../../utils/getModifiedProps';
 import { logSeriesErrorMessage } from '../../utils/warnings';
 
-class Series extends Component {
+class Series extends PureComponent {
 
   static propTypes = {
     id: PropTypes.oneOfType([ PropTypes.string, PropTypes.func ]).isRequired,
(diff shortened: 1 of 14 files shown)


packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -47,10 +43,7 @@ class Series extends Component {
 
     let needsRedraw = false;
     // Using setData is more performant than update
-    if (isImmutable(data) && immutableEqual(data, prevProps.data) === false) {
-      this.series.setData(data.toJS(), false);
```
whawker: I feel uneasy about dropping support for Immutable, I can't see how these few lines effect the bundle size that much, and Immutable is quite likely to be used in a React app.

anajavi: It's not about bundle size, it's only a couple of kB. I just thought that making equality checks is unnecessary. 

I can bring it back with immutable 4?

What about isEqual checks? I find it more inline with other libraries when doing just shallow compare?

anajavi: Actually if we keep immutable and isEqual, there is no need to make major release if Highcharts 7 does not break things.

All the new features and lodash changes can be made within version 3.

Might make more sense actually, as most React libraries seem to target their breaking changes to React 16.7.

I am not against that all. What do you think?

anajavi:
Postponed.

Issue #182, opened 2018-11-05
Update annotations?


knod:
#### Bug or Feature Request?
Not sure. There may already be a way to do this.

#### Description
First off, I can't figure out how to have just one annotation for the chart, not for every series, so ignore that part of the example for now. That aside, how do I update an annotation that should be based off of a dynamic value? One that's handed in as a prop?

#### How to reproduce
I think the examples will show it better.

#### Live demo demonstrating bug

This has a vertical line. There's a button at the bottom. When you press that button, a random number is generated and it's sent to the chart. The vertical line's x-position should then be at that x location, but it always stays in the same place:

- Version with just a function with an object: https://stackblitz.com/edit/base-chart-example-h9cxt7
- Version with lifecycle method for updating annotation: https://stackblitz.com/edit/base-chart-example-exy6bu
- Version with `onAddSeries()`: https://stackblitz.com/edit/base-chart-example-gq4kck

I took my cue on annotations from https://github.com/whawker/react-jsx-highcharts/issues/136.

#### Versions
* React JSX Highcharts version: ^3.3.0
* Highcharts version: ^6.2.0
* React version: 16.3.2

anajavi:
Here's a quick & dirty example of updating annotations:
https://stackblitz.com/edit/base-chart-example-ejsoey?file=MyChart.js

The MyChart component is itself managing the annotations and keeping them in this.chartAnnotations.

knod:
@anajavi : Hmm, interesting. How can we get it to appear when the component is mounted? Also, is `ant` even needed anymore?

knod:
Forget that first question. We can just call `this.updateAnnotations()` in `getChart()`.

knod:
Is there a way to create an annotation without `.addAnnotation()`?

knod:
So it occurred to me that what you could do is create a data point just for the annotation and then attach the annotation to it. I started simple by removing the annotation stuff and made a new line: `<LineSeries name="X" data={[this.props.x]} />`. Thing is, it doesn't update its position when x changes either. How do I get a line to change when the value of its props change?

knod:
The chart's `.componentDidUpdate()` does run: https://stackblitz.com/edit/base-chart-example-6yc4is?file=index.js

knod:
That same link should show a plot line updating now. That's all I really need for this problem, but I'm going to need to be able to update line positions. Is there a solution for that?

knod:
[...]

PR #190, opened 2018-11-14
Add Annotation component

anajavi:
Adds basic annotation support.

### DIFF
--- a/packages/react-jsx-highcharts/src/utils/warnings.js
+++ b/packages/react-jsx-highcharts/src/utils/warnings.js
@@ -11,6 +11,7 @@ const descCommentCSS = descCSS('#808080') + 'font-style:italic;';
 const descNewLine = 'font-size: 1px; margin-right: 100%;';
 
 const moduleToImportPath = {
+  annotations: 'modules/annotations',
   more: 'highcharts-more',
   threeD: 'highcharts-3d',
   bullet: 'modules/bullet',
@@ -31,6 +32,7 @@ const moduleToImportPath = {
 };
 
 const moduleToVarName = {
+  annotations: 'addAnnotations',
   more: 'addHighchartsMore',
   threeD: 'addHighcharts3DModule',
   bullet: 'addBulletModule',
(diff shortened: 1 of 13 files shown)


whawker:
This all seems reasonable, would be worth adding an example with annotations I think.

packages/react-jsx-highcharts/src/utils/warnings.js
```diff
@@ -51,6 +53,7 @@ const moduleToVarName = {
 };
 
 const moduleToSeriesTypes = {
+  annotations: ['annotations'],
```
whawker: Maybe worth renaming this var to `moduleToFeatureMap` ?

anajavi: I couldn't figure good name for it, so thank you!

anajavi:
Added an example.

anajavi:
alpha release or wait for highcharts 7?

anajavi:
Maybe @knod from #182 would like to test alpha?

whawker:
I would suggest an alpha release, are you happy with the stability of the Lodash removal stuff you've been working on?

anajavi:
I see you already fixed null reference from one of them, sorry about that.

I think I will check them one more time.

knod:
Sure, if it'd be helpful I can try it out. I can just clone this lib and try it out from there if it's not in an npm-able alpha yet.

knod:
Sure, if it'd be helpful I can try it out. I can just clone this lib and try it out from there if it's not in an npm-able alpha yet. Got any specific questions you want me to focus on?

anajavi:
@whawker I added some tests for the lodash changes. Now I think it is ready to be released.

@knod there is an example at https://github.com/whawker/react-jsx-highcharts/blob/master/examples/AreaWithAnnotations/App.js
I got no specific questions, just want to know if it is easier to work with.

anajavi:
Works for me. The first render is now very fast.

PR #204, opened 2019-02-17
Export Consumer and Provider context in index.js

msteinmn:
Creates the ability to provide a custom withHighcharts HOC. I need this because I am using react-docgen to generate documentation for the Highcharts components. react-docgen creates an error with the existing HOC: Error: No suitable component definition found.

It would be awesome if you could merge this simple addition to the export list. Would help me not to have to fork / republish.

### DIFF
--- a/packages/react-jsx-highcharts/src/index.js
+++ b/packages/react-jsx-highcharts/src/index.js
@@ -69,7 +69,7 @@ export const WindBarbSeries = withSeriesType('WindBarb');
 export const XRangeSeries = withSeriesType('XRange');
 
 // Providers
-export { Consumer, Provider } from './components/HighchartsContext';
+export { Consumer as HighchartsConsumer, Provider as HighchartsProvider } from './components/HighchartsContext';
 export { default as provideHighcharts }  from './components/HighchartsProvider';
 export { default as provideChart }  from './components/ChartProvider';
 export { default as provideAxis }  from './components/AxisProvider';


packages/react-jsx-highcharts/src/index.js
```diff
@@ -69,6 +69,7 @@ export const WindBarbSeries = withSeriesType('WindBarb');
 export const XRangeSeries = withSeriesType('XRange');
 
 // Providers
+export { Consumer, Provider } from './components/HighchartsContext';
```
whawker: This seems like a sensible idea, I'm just wondering if perhaps `Consumer` and `Provider` are a bit too generic.

```suggestion
export { Consumer as HighchartsConsumer, Provider as HighchartsProvider } from './components/HighchartsContext';
```

whawker: We also use context for Charts, Axes and Series - does the context for those need exposing too?

msteinmn:
Thanks for reviewing it. I have not (yet) come across the need for other Contexts, but exporting them too would likely be a good idea.

whawker:
Cool in that case I would suggest...

```js
export { Consumer as HighchartsConsumer, Provider as HighchartsProvider } from './components/HighchartsContext';
export { Consumer as ChartConsumer, Provider as ChartProvider } from './components/ChartContext';
export { Consumer as AxisConsumer, Provider as AxisProvider } from './components/AxisContext';
export { Consumer as SeriesConsumer, Provider as SeriesProvider } from './components/SeriesContext';
```

anajavi:
How about exporting the context object? react-redux seems to be doing that:
https://github.com/reduxjs/react-redux/blob/fac9ad19d17000f5b9fdd52fc9843225b807845b/src/index.js#L3

whawker:
Interesting, I take it that works with your react-docgen requirement?

msteinmn:
Not sure what the difference is. As long as I can create a custom HOC withHighcharts outside the package. I like your second (extended) suggestion.

whawker:
Just out of curiosity, what does your custom `withHighcharts` HOC look like?

msteinmn:
`
import React from 'react';
import { Provider } from 'react-jsx-highcharts';

// This is a HOC function.
// It takes a component...
export default (Highcharts) => (Component) => {
  // ...and returns another component...
  return function HighchartsWrappedComponent(props) {
    // ... and renders the wrapped component with the context theme!
    // Notice that we pass through any additional props as well
    return (
      <Provider value={Highcharts}>
        <Component {...props} />
      </Provider>
    );
  };
};
`
It is practically the same except for the way it is invoked:
export default withHighcharts(Highcharts)(Chart);

I have this working on a forked (local) version of react-jsx-highcharts.

whawker:
Thanks. I will merge this, then make a small modification that follow the pattern ReactRedux use. I'll get back to you when it's done.

msteinmn:
Awesome, thanks.

whawker:
[Here](https://github.com/whawker/react-jsx-highcharts/commit/912bb852d55f2dae75f189dddd64b3f5ae795474) are the changes I made.

From your example you should be able to

```jsx
import React from 'react';
import { HighchartsContext } from 'react-jsx-highcharts';

// This is a HOC function.
// It takes a component...
export default (Highcharts) => (Component) => {
  // ...and returns another component...
  return function HighchartsWrappedComponent(props) {
    // ... and renders the wrapped component with the context theme!
    // Notice that we pass through any additional props as well
    return (
      <HighchartsContext.Provider value={Highcharts}>
        <Component {...props} />
      </HighchartsContext.Provider>
    );
  };
};
```

I'll go ahead and publish this now.

whawker:
Published

* `react-jsx-highcharts` 3.5.0
* `react-jsx-highstock` 3.5.0
* `react-jsx-highmaps` 1.2.0

PR #212, opened 2019-06-12
Add immutable 4 tests

anajavi:
Do immutable tests with both immutable@3 and @4.

This is preparation for replacing is-immutable and immutable-is with immutable 4 (still RC).

Package aliasing is not supported in npm < 6.9.0, so this introduces it as a dependency for development. npm 6.9.0 is bundled with node 10.16.0.
```json
"immutable3": "npm:immutable@^3.8.1",
```
Can this be merged or is older npm needed somewhere?

Congratulations for the new family member!


whawker:
Just a late thought (sorry) we should probably add an [`engines`](https://docs.npmjs.com/files/package.json#engines) definition for npm to the `package.json`?

anajavi:
That might have some unexpected results for people installing from npm? As the aliased dependency is only in devDependencies, it should not matter to library users.

whawker:
Ah good point, fair enough. If I every get round to writing a `CONTRIBUTING.md, I'll add a note in there

anajavi:
Immutable 3 tests pass also when replacing is-immutable and immutable-is with immutable 4.

bundle size:
before: 49190
after: 48691

Not much.

Issue #209, opened 2019-05-12
Charts are visibly refreshing when updated


pinkynrg:
On version 3.3 this is not happening.

From version 3.4 it almost looks like that updated charts disappear and reappear in a blink of an eye, but it is visible.

kMeillet:
I got the exact same problem.

I'm using Highstock with react-jsx-highcharts.

Navigator is enabled (set to 1mn).

LineSeries work well.

But ColumnSeries refresh the whole chart.

whawker:
Apologies for the late reply, my son was born 5 weeks ago, and I haven't had time to look after my open source stuff. Looking into this now.

whawker:
@pinkynrg can you please give more information on how to reproduce what you are seeing. Ideally a CodeSandbox/Stackblitz reproducing the problem, or even information on which series types/browser you are using.

@pinkynrg and @kMeillet by any chance, is the issue you're seeing on Firefox or Safari? Just wondering if it's related to this bug https://github.com/highcharts/highcharts/issues/9245

pinkynrg:
Right now I don't have a lot of time but I can tell you right now that the
issue is happening on Chrome too.

whawker:
How many data points do the series on your charts general have? Are you using Highcharts boost module?

pinkynrg:
Generally not more than 300 points per chart so no, I'm not using Highcharts boost, since I don't think the data load is high enough to justify its use.

whawker:
Would either of you be able to post an example of your code? It's difficult to debug in the dark here.

jhartling:
Here's a reproduction: https://stackblitz.com/edit/base-chart-example-h2gjjk

whawker:
Thanks @jhartling, I have managed to reproduce on vanilla Highcharts, so I have raised the follow issue https://github.com/highcharts/highcharts/issues/10883

PR #213, opened 2019-06-12
Add Highcharts 7.1 support

whawker:
Also fixes a few outstanding bugs.

Most notably #209, where we now use `requestAnimationFrame` for debouncing chart redraws, (instead of lodash `debounce`). 

This is necessary as Highcharts [destroys axes when they are updated](https://github.com/highcharts/highcharts/blob/master/js/parts/Dynamics.js#L1130), so we need to redraw it by the next animation frame, otherwise we see the white flash.

This seems to be the best compromise between limiting unnecessary redraws, and prevent white flashes.


whawker:
Thinking I will gitignore the `bundle.js` and `index.html` example files on the master branch, but keep them on the gh-pages branch. Seems no value in having them on master.

anajavi:
I like the idea of using requestAnimationFrame instead of setTimeout.

The code looks fine to me, but I'd like to check how it behaves in profiler. Hopefully by tomorrow.

If you are preparing for a release I could also update some dependencies. Notably switching from create-react-context to https://github.com/StringEpsilon/mini-create-react-context which is used by react-router nowadays

whawker:
Sounds like a good idea. Yep, I'm looking to push out a release soon - have been neglecting this project recently, but I don't want people to think it's dead!

anajavi:
Quick profiling did not reveal anything special, so I say it is good to go.

Great work!

PR #216, opened 2019-06-17
Rewrite using hooks

anajavi:
This moves getChart() function from provideChart to BaseChart. I left old context values (chart, chartType) in place as ChartContext is now exported and maybe used outside this code.

Goals:
- getChart is cached in BaseChart state, so same instance is always provided to components, [preventing react rerenders](https://reactjs.org/docs/context.html#caveats)
- easier to [drop some HOC's with newer react version](https://reactjs.org/docs/context.html#classcontexttype)
- easier to use with hooks

Same should be done to HighchartsContext, AxisContext and SeriesContext, but I thought to ask:
1. is there is any known pitfalls with this approach @whawker ?
2. should I try to prevent any breaking changes, or are we ready for a major version bump?


whawker:
Nice observation on the same instance thing, that should be a performance win. The code so far looks good to me too.

On the `contextType` thing, we have components such as `Series` that interact with multiple contexts ([Highcharts](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/Series/Series.js#L39), [Chart](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/Series/Series.js#L107) and [Axis](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/Series/Series.js#L99)), and it looks to me like `contextType` means we could only interact with one?

It seems like render props/HOCs seem to be the expected pattern if dealing with [multiple contexts](https://reactjs.org/docs/context.html#consuming-multiple-contexts)? 

If we decided to drop HOCs, that would definitely be a major version bump in my eyes.

anajavi:
With contextType only one context can be consumed. With hooks multiple can be used. Rewriting with hooks requires bumping to React 16.8 though.

I checked the code, and actually the ChartContext is the only context that can be changed without changing the value that Provider is setting to context. I would consider these as breaking changes because of the exported Context objects.

To continue this PR the options seem to be:
1. Merge current changes and let other contexts be like they are. Can be released in a minor version
2. Change other contexts and release a major version
3. Bump React version, rewrite with hooks. Release a major version
4. Do nothing, close this PR

whawker:
Having thought about this for a while now, I think we should definitely do option 1 in the short term, but longer term also begin work on a new major version rewriting the project using hooks (would be good for me personally, as I really need to learn them!).

anajavi:
That sounds good. I shall then merge this to master.

If I find the time I can make a branch that fixes the rest of contexts and we can base the hooks version on that branch. It is then possible to choose which version to use for the next major version.

anajavi:
Actually when trying to fix other problems reported by eslint-plugin-perf, I realised that these Contexts can also be fixed with https://github.com/alexreardon/memoize-one

So I will hold this PR a while, maybe we can fix these without breaking changes.

anajavi:
The context provide functions can indeed be fixed with memoize-one. However, the wrapped component needs to be Class not function:

anajavi:
I have tried making the  context changes, but it seems that it is easier to go straight to using hooks.

@whawker can you tell me whats the reason for DelayRender on:
https://github.com/whawker/react-jsx-highcharts/blob/efe950bf10069adc0554a6b114201c54529f925a/packages/react-jsx-highcharts/src/components/AxisProvider/index.js#L18
Does Highcharts require that axis is rendered before it can be accessed?

Anyway, I think that this PR can be merged now.

whawker:
[DelayRender](https://github.com/whawker/react-jsx-highcharts/search?q=delayrender&unscoped_q=delayrender) is required for axes, series and plotlines, as they can't be added until the parent object exists. 

Axes can't be added till the chart exists and series can't be added until both the chart and [parent axis](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/Series/Series.js#L100) (or [`coll`](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/AxisProvider/index.js#L32)) exists. I _think_ it also helps in the remove case, i.e. avoiding unnecessary work removing series when the parent axis was unmounted

anajavi:
Status update:

I decided to use this branch for hooks rewrite as #218 took care of the immediate problems.

The bundle size goes down with every component rewritten. It also "feels" faster, but that might be placebo effect.

The AxisContext usage needs to be rewritten completely, which is kind of hairy.

Lots of work remaining still..

whawker:
It's looking really good! I've only had minimal exposure to hooks, so it's interesting seeing them being applied to a codebase I know really well.

I assume this will form a new major version as previously discussed?

anajavi:
> I assume this will form a new major version as previously discussed?

Yes it will. Breaking changes are unavoidable with this.

Some other breaking changes I'd like to make, if you don't oppose them:
- Make series shallow compare data instead of isEqual
- Drop support for immutable as it is [unmaintained](https://github.com/immutable-js/immutable-js/issues/1689)
- Drop Highcharts < 7 (or < 8 depending on how long this PR takes)
- Remove FlagSeries as it is now called FlagsSeries
- Make Series throw if it requires Axis but it is not provided with one instead of returning null
- Require modern browser (what is the baseline for modern browser??)

Other considerations if not too tricky:
- Importing this library with es2015 imports from node 12 (https://nodejs.org/api/esm.html)

I am pretty sure that this will take some time to make though.

anajavi:
Status:
- most classes are rewritten to functions
- bundle size has gone down over 10kB
- react devtools already shows much nicer component tree than before
- performance should be better as most components are memo'ed now
- there will be bugs

I have a question though:
The axis context rewrite is blocking on converting PlotLineLabel and PlotBandLabel to functional components. Those two are quite complex.
I think that all their functionality can be reproduced by passing options to PlotLine and PlotBand. I'd like to remove both as they are quite brittle.

@whawker what do you think?

whawker:
What part adds the complexity? Is it [this line](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/PlotBand/PlotBandLabel.js#L67)? (or that function?) It is definitely a hack I wish we didn't have.

I'd prefer to keep the Plot*Label components if at all possible. I personally think one the strengths of React JSX Highcharts is that it (generally) avoids passing round objects as configuration, by providing components instead.

anajavi:
> What part adds the complexity? Is it [this line](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/PlotBand/PlotBandLabel.js#L67)? (or that function?) It is definitely a hack I wish we didn't have.

Yes. That part of the labels kind of fights against the way React likes to do things. And the parent component needing to add and remove plotlines on every render does not help with the complexity.

> I'd prefer to keep the Plot*Label components if at all possible. I personally think one the strengths of React JSX Highcharts is that it (generally) avoids passing round objects as configuration, by providing components instead.

Maybe the Plot*Label components can be rewritten to control the parent component state? But then Plot*Label's need to be always wrapped with PlotLine or PlotBand. Is that fair compromise?

whawker:
Yes it's a fair compromise, I would always expect a Plot(Band/Line)Label to be rendered inside a `PlotBand`/`PlotLine` component.

Just out of curiosity, would the `useContext` hook help here at all?

anajavi:
> Yes it's a fair compromise, I would always expect a Plot(Band/Line)Label to be rendered inside a `PlotBand`/`PlotLine` component.

Let's do it that way then.

> Just out of curiosity, would the `useContext` hook help here at all?

I think that the simplest way would be to create PlotLineContext in PlotLine and use it via useContext in PlotLineLabel. That should make the requestAnimationFrame unnecessary in PlotLineLabel.

I actually already have custom useDelay hook, but testing it seems to be quite a nightmare:
```javascript
import { useEffect, useState, useRef } from 'react';


export default function useDelay () {
  const [delayPassed, setDelayPassed] = useState(false);
  const delayTimeout = useRef(null);
  const unmounted = useRef(false);

  useEffect(() => {
    delayTimeout.current = window.requestAnimationFrame(() => {
      if (unmounted.current === false) {
        setDelayPassed(true);
      }
      delayTimeout.current = null;
    });
    return () => {
      unmounted.current = true;
      if(delayTimeout.current !== null) {
        window.cancelAnimationFrame(delayTimeout.current);
      }
    }
  }, []);
  return delayPassed;
}

```

whawker:
Here's a [test for `useInterval`](https://github.com/donavon/use-interval/blob/master/__tests__/index.test.js) that might help?

anajavi:
> Here's a [test for `useInterval`](https://github.com/donavon/use-interval/blob/master/__tests__/index.test.js) that might help?

Actually the problem is testing components that depend on useDelay, like DelayRender:
```javascript
import React from 'react';
import useDelay from '../UseDelay';

const DelayRender = ({children}) => {
  const render = useDelay();

  if (!render) return null;

  return children;
}

export default DelayRender;
```

The state change in requestAnimationFrame causes all the integration tests to warn about it, although they don't fail:
```
      Warning: An update to DelayRender inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
```

The warning can be probably prevented by mocking requestAnimationFrame and wrapping the callback in act(() => { callback }), but it spreads to several tests then. Doable, but very dirty and annoying.

Also, I found this library for hooks testing: https://www.npmjs.com/package/@testing-library/react-hooks

packages/react-jsx-highcharts/src/components/Annotation/Annotation.js
```diff
@@ -1,88 +1,62 @@
...
+import useChart from '../UseChart';
 
-class Annotation extends Component {
+const Annotation = memo((props) => {
+  const { id = uuid, children, ...rest } = props;
 
-  static propTypes = {
-    getChart: PropTypes.func.isRequired // Provided by ChartProvider
-  };
+  const { getChart } = useChart();
```
whawker: It might defeat the purpose of memoing everything, but is it possible for `useChart` to return [these methods](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/ChartProvider/index.js#L15-L25) directly? Instead of having to call `getChart()` to access them?

Reason being, a lot of our components use `chart.update({ name: { ...opts } })` and it would be nice if we could do

```js
const { update } = useChart()
update({ legend: { ...modifiedProps } })
```

I'm beginning to realise now (as you originally suggested) how hooks allow us to drop our HOC components. I can see the value of potentially dropping "context" completely and storing the current chart instance inside `useChart` via a `useRef`?

anajavi: > It might defeat the purpose of memoing everything, but is it possible for `useChart` to return [these methods](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/ChartProvider/index.js#L15-L25) directly? Instead of having to call `getChart()` to access them?
> 
> Reason being, a lot of our components use `chart.update({ name: { ...opts } })` and it would be nice if we could do
> 
> ```js
> const { update } = useChart()
> update({ legend: { ...modifiedProps } })
> ```

I think that is a good idea. Probably easiest to do when all provideChart calls are removed (soon).

> 
> I'm beginning to realise now (as you originally suggested) how hooks allow us to drop our HOC components. I can see the value of potentially dropping "context" completely and storing the current chart instance inside `useChart` via a `useRef`?

useChart still needs to get the chart instance from somewhere, so context is needed for that.

packages/react-jsx-highcharts/src/components/PlotLine/PlotLine.js
```diff
@@ -1,85 +1,53 @@
...
-
-    return (
-      <Hidden>
-        {lineChildren}
-      </Hidden>
-    );
-  }
-}
-
+    axis.addPlotLine(opts);
```
anajavi: PlotLine and PlotBand are currently broken. Some kind of change tracking is needed as these have children.

anajavi: I think PlotLine, PlotBand and Annotation could use shared lifecycle like useChartUpdate is for Title, Legend etc.

packages/react-jsx-highcharts/test/components/Series/SeriesTypes.integration.spec.js
```diff
@@ -92,6 +104,9 @@ Object.keys(all).filter(name => /^[A-Z].*Series$/.test(name)).forEach((seriesNam
         };
         const WithComponent = withHighcharts(Component, Highcharts);
         const renderedChart = renderIntoDocument(<WithComponent />);
+        act(() => {
+          jest.runAllTimers();
```
anajavi: Using fake timers slows down these tests 3-4 times = 30s. Hopefully this can be fixed somehow.

whawker:
Sorry for the lack of responses, I had a family emergency on Thursday morning. So haven't been able to look through any of the commits. Apologies

anajavi:
> Sorry for the lack of responses, I had a family emergency on Thursday morning. So haven't been able to look through any of the commits. Apologies

I hope your family is alright now. And don't apologise, this code certainly can wait.

anajavi:
Status:
- [ ] Make series shallow compare data instead of isEqual
  _Shall we make this change?_
- [ ]  Drop support for immutable as it is [unmaintained](https://github.com/immutable-js/immutable-js/issues/1689)
   _Shall we make this change?_
- [X]  Drop Highcharts < 7 (or < 8 depending on how long this PR takes)
- [X] Remove FlagSeries as it is now called FlagsSeries
- [ ]  Make Series throw if it requires Axis but it is not provided with one instead of returning null
  _The Axis can be delayed, hard to know when it is proper to throw. Won't do this_
- [ ]  Require modern browser (what is the baseline for modern browser??)
  _Dropped IE < 11, as I can't find suitable baseline or example project doing this_
- [X] Rewrite classes to functions
- [X] Drop provide functions
- [ ] Remove defaultProps as they are deprecated in functional components
- [ ] It might be possible to replace useDelay with useLayoutEffect or useEffect. This would simplify things.
- [ ] Extract Series lifecycle to a hook. Use the hook in withSeriesType instead of passing props to Series.
  _Should we do this?_
- [ ] Extract Axis lifecycle to a hook. Use the hook in XAxis, YAxis and ZAxis instead of passing props to Axis.
  _Should we do this?_
- [ ] Are we happy with the naming of things like hooks, eg.?
- [ ] Write tests for the custom hooks
- [ ] Rewrite react-jsx-highstock
- [ ] Rewrite react-jsx-highmaps
- [ ] Decide what to do with react-jsx-highstock-datepickers. The published version still has peerDependency for Highcharts 6. [The weekly downloads are quite low](https://www.npmjs.com/package/react-jsx-highstock-datepickers). And it needs new version of react-day-picker, which is quite a bit of work.

Other:
- [ ] Importing this library with es2015 imports from node 12 (https://nodejs.org/api/esm.html)
  _This proved to be too complex at this time. Webpack behaves differently when input is .mjs file for example. I suppose there will be solutions for this after node 12 reaches LTS_

anajavi:
In order to finalize this, these decisions should be made:
1. Use shallow compare data instead of isEqual
This means that getModifiedProps, Series and BaseChart would use Object.is to check for equality instead of checking for deep equality with isEqual. It should be quite a bit faster with larger datasets.
[Object.is requires a polyfill in IE11, but it is quite short code.](https://github.com/facebook/react/blob/master/packages/shared/objectIs.js)

2. Drop support for immutable as it is [unmaintained](https://github.com/immutable-js/immutable-js/issues/1689)
Very slight decrease in bundle size and simpler Series component are the benefits of this. Most new projects seem to be using Immer instead of Immutable-js.

3. Decide what to do with react-jsx-highstock-datepickers.
The published version still has peerDependency for Highcharts 6. [The weekly downloads are quite low](https://www.npmjs.com/package/react-jsx-highstock-datepickers).
A new react-day-picker is required to remove deprecated React lifecycle methods. [The new react-day-picker also drops moment.js from dependencies](https://github.com/gpbl/react-day-picker/blob/master/CHANGELOG.md), so I think the whole datepickers component should be refactored.
Should we keep the react-jsx-highstock-datepickers or deprecate it in npm?

4. Support only modern browsers
I could not find any examples doing this properly. Maybe we should just drop IE < 11 and require Object and Array polyfills for IE? I am pretty sure most people polyfill IE fully as React itself requires Map and Set polyfills.
Babel transpile without IE support would reduce the bundle size some 30%. But then again, it is easier to ship src folder in npm package and let people decide the baseline themselves.

Also a alpha release would be nice for testing purposes. We could create a github issue for alpha feedback with changelog.

@whawker what do you think?

whawker:
Hi @anajavi sorry for the lack of input, due to family stuff as I said. I do massively appreciate the work you're doing here!

1. That sounds reasonable, I can run lots of scenarios during our alpha tests so we can raise any issues and decide whether they're acceptable.
2. Yep a fair point, should we consider supporting immer instead? We could potentially have a separate package (in the future) i.e. `react-jsx-highcharts-immer-series` to reduce the bundle size for everyone else?
3. Yeah we can bin this, I had lofty dreams of a React JSX Highcharts component ecosystem, where people made their own plugins etc, but this was misguided.
4. Sure we'll just make it clear on the readme what polyfills are required (and facepalm every time someone raises an IE10 issue)

I can publish an alpha now, if you're happy?

Thanks again for all your hard work! Incredible!

anajavi:
> Hi @anajavi sorry for the lack of input, due to family stuff as I said. I do massively appreciate the work you're doing here!

No code is more important than family.

> 1. That sounds reasonable, I can run lots of scenarios during our alpha tests so we can raise any issues and decide whether they're acceptable.

Great! I will do this then.

> 2. Yep a fair point, should we consider supporting immer instead? We could potentially have a separate package (in the future) i.e. `react-jsx-highcharts-immer-series` to reduce the bundle size for everyone else?

Immer seems to be using plain javascript object and arrays, so no support should be needed.
I will remove the immutable support.

> 3. Yeah we can bin this, I had lofty dreams of a React JSX Highcharts component ecosystem, where people made their own plugins etc, but this was misguided.

Alright. I guess it should be then deprecated with:
```
npm deprecate react-jsx-highstock-datepickers "<message>"
```
That can wait for the final release though.

> 4. Sure we'll just make it clear on the readme what polyfills are required (and facepalm every time someone raises an IE10 issue)

Something like this would let us use most of the functions: "Using v4.0.0 requires Object and Array polyfilled to at least es2015 level. If you are supporting IE11, please add core-js polyfills to your code."

> I can publish an alpha now, if you're happy?

I can do 1. and 2. first, so probably tomorrow if that is alright with you?

whawker:
Yep that's fine, I'll also try and get back to you on those questions you raised 7 days ago regarding the original implementation.

anajavi:
Immutable removed and isEqual replaced with Object.is. Was surprisingly easy.

So now it is ready for alpha release.

anajavi:
Release notes draft:

# Version 4 release notes

Version 4 is a rewrite of the libraries with React hooks.

What's new in version 4:
- React rerender performance should be improved
- Full bundle size is around 25% smaller

Most of the time upgrade to version 4 should go without changes to the code, but there are some breaking changes. If you are having problems with upgrade, be sure to raise a issue in github.

## Breaking changes

- Require React 16.8.6
- Require Highcharts 7.2.0
- Drop support for Immutable-js
- Props are now shallow compared instead of version 3's deep equal checks.
Be sure you are not placing new objects or arrays to props, otherwise the components will rerender needlessly
- Minimum supported version of IE is 11
Also IE needs to be polyfilled to at least es2015 level, which you are probably already doing with core-js. If you are using create-react-app, please see https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill#user-content-polyfilling-other-language-features

These should not affect many users:
- provide methods are replaced with hooks:
  - provideHighcharts --> useHighcharts
  - provideChart --> useChart
  - provideAxis --> useAxis
  - provideSeries --> useSeries
- Values in Contexts are changed from functions to objects

## New features
- Caption component for setting chart caption
```jsx
<Caption>
  This is a caption.
</Caption>
```
- Hooks for extending the library

whawker:
@anajavi I have published

* `react-jsx-highcharts@4.0.0-alpha.1`
* `react-jsx-highstock@4.0.0-alpha.1`
* `react-jsx-highmaps@2.0.0-alpha.1`

All using the `next` tag on NPM

packages/react-jsx-highcharts/src/components/Hidden/Hidden.js
```diff
@@ -1,20 +1,7 @@
...
-  render () {
-    const { children } = this.props;
-    if (!children) return null;
-
-    return [
-      children
-    ];
-  }
-}
+const Hidden = ({ children = null}) => (
```
anajavi: What is the purpose of <Hidden>? Could it be replaced with <React.Fragment>?

anajavi:
The withHighcharts is the last remaining HOC in codebase.

I'd like to add new component to do it's duties, perhaps something like this:
```jsx
<WithHighcharts Highcharts={Highcharts}>
  <HighchartsChart .... />
</WithHighcharts>
```
Can't think of a better name for it though.

For easy migration, withHighcharts should be kept for v.4, but a deprecation message could be added to it.

What do you think about it?

packages/react-jsx-highcharts/src/components/Hidden/Hidden.js
whawker: It can yes, it was a legacy component from back before React.Fragment existed.

packages/react-jsx-highstock/src/components/RangeSelector/RangeSelectorInner.js
```diff
@@ -1,103 +1,85 @@
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { attempt } from 'lodash-es';
-import { Hidden, getModifiedProps, provideAxis } from 'react-jsx-highcharts';
+import { useHighcharts, useAxis, useChart, useModifiedProps } from 'react-jsx-highcharts';
 
-class RangeSelectorInner extends Component {
+const RangeSelectorInner = ({ enabled = true, axisId, children, ...restProps }) => {
```
anajavi: note: I think that RangeSelector and RangeSelectorInner could now be merged into single component

whawker: I suspect you are correct - I believe the original reason for this was this need to [provide an axis](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highstock/src/components/RangeSelector/RangeSelectorInner.js#L100), but it looks as though with the hooks this shouldn't be necessary

packages/react-jsx-highstock/src/components/RangeSelector/RangeSelector.js
```diff
@@ -1,8 +1,85 @@
...
 
-export default RangeSelector;
+
+  useEffect(() => {
+    if (!axis) return;
+
+    // Workaround inferred from http://jsfiddle.net/x40me94t/2/
+    const chartObj = chart.object;
+    chartObj.options.rangeSelector.enabled = true;
+    Highcharts.fireEvent(chartObj, 'init'); // Pre Highcharts 6.1
```
whawker: We can get rid of this, as v4 will only support > Highcharts 7.2

anajavi: Can you make that change? I don't really understand what is happening there :)

whawker: Will do, just about to publish a new alpha too. Then I think we'll merge this to master, we can fix any new issues raised on there.

packages/react-jsx-highcharts/src/components/PlotBandLine/usePlotBandLineLifecycle.js
```diff
@@ -23,17 +23,13 @@ export default function usePlotBandLine(props, plotType) {
         id: myId,
         ...rest
       };
-
-      if(plotbandline) axis.removePlotBandOrLine(idRef.current);
       setPlotbandline(axis.addPlotBandOrLine(opts, plotType));
```
anajavi: Something strange with axis.addPlotBandOrLine. It does seem to return undefined, but I've seen PlotLineLabel rendering correctly, which depends on it returning object.

whawker: I suspect this is the bit causing issue, the plot line/band render method [can return undefined](https://github.com/highcharts/highcharts/blob/a188b87b3d2b4f386672ee48e547110e8e61b2a2/js/parts/PlotLineOrBand.js#L127) if `isLine` and `isBand` are false.

Which are defined as
```js
isBand = defined(from) && defined(to)
isLine = defined(value)
```

So potentially `value` is undefined when creating a Plot Line, or either `from` or `to` is undefined when creating a Plot Band?

anajavi: I am actually testing against this: https://stackblitz.com/edit/base-chart-example-olzxfe?file=MyChart.js

The plugin code might actually be hiding the returned object, because the plotline value is actually set.

I am not familiar with highcharts plugins, but I must check this tomorrow without the plugin.

anajavi: It was the plugin code. It is not returning on proceed.apply(....

anajavi: I think I have done all things that I had in mind before next alpha.

whawker: Cool, happy for me to merge this PR to master?

anajavi: More than happy!

whawker:
Not much else I can say, incredible work on this @anajavi!

PR #218, opened 2019-06-25
Add eslint-plugin-react-perf and fix errors

anajavi:
I added eslint-plugin-react-perf to detect performance problems in eslint.

It currently errors out, but it could prevent problems like #216.

@whawker do you think these rules should be turned into warnings and merge this to master? After fixing these the rules should probably be turned to errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,7 @@
 {
   "root": true,
   "parser": "babel-eslint",
-  "plugins": ["react", "import"],
+  "plugins": ["react", "import", "react-perf"],
   "parserOptions": {
     "ecmaVersion": 2018,
     "sourceType": "module",
@@ -16,7 +16,8 @@
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended",
-    "plugin:import/errors"
+    "plugin:import/errors",
+    "plugin:react-perf/recommended"
   ],
   "rules": {
     "no-unused-vars": [
--- a/packages/react-jsx-highcharts/package-lock.json
+++ b/packages/react-jsx-highcharts/package-lock.json
@@ -3319,6 +3319,12 @@
         }
       }
     },
+    "eslint-plugin-react-perf": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-react-perf/-/eslint-plugin-react-perf-3.2.0.tgz",
+      "integrity": "sha512-PnAsJPwCmoI4CBtF/Uk0v1dN4zvpWC3fO5BzG9O+q/c+6hXv9cFzgXht+CJzZaC61VuYBpjWdOs9ksBN6Py1iw==",
+      "dev": true
+    },
     "eslint-scope": {
       "version": "3.7.1",
       "resolved": "https://registry.npmjs.org/eslint-scope/-/eslint-scope-3.7.1.tgz",
--- a/packages/react-jsx-highcharts/package-lock.json
+++ b/packages/react-jsx-highcharts/package-lock.json
@@ -6159,6 +6159,11 @@
         }
       }
     },
+    "memoize-one": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/memoize-one/-/memoize-one-5.1.1.tgz",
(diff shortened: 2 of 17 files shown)


whawker:
Good idea;

When you say "errors out", I don't know if you meant fails completely, here are the errors I can see.

Perhaps we should look to fix these errors as part of this PR?

Highcharts
```
/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/AxisProvider/index.js
  29:29  error    JSX attribute values should not contain functions created in the same scope  react-perf/jsx-no-new-function-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
  122:28  error    JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/ChartProvider/index.js
  19:28  error  JSX attribute values should not contain functions created in the same scope  react-perf/jsx-no-new-function-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/Highcharts3dChart/Highcharts3dChart.js
  14:19  error  JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  17:19  error  JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/HighchartsProvider/index.js
  15:48  error  JSX attribute values should not contain functions created in the same scope  react-perf/jsx-no-new-function-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/HighchartsSparkline/HighchartsSparkline.js
  54:24   error  JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  71:24   error  JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  71:97   error  JSX attribute values should not contain Arrays created in the same scope   react-perf/jsx-no-new-array-as-prop
  73:39   error  JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  73:112  error  JSX attribute values should not contain Arrays created in the same scope   react-perf/jsx-no-new-array-as-prop

/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highcharts/src/components/SeriesProvider/index.js
  26:31  error    JSX attribute values should not contain functions created in the same scope  react-perf/jsx-no-new-function-as-prop
```

Highmaps
```
/Users/williamhawker/dev/react-jsx-highcharts/packages/react-jsx-highmaps/src/components/HighchartsMapChart/HighchartsMapChart.js
  36:16  error    JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  37:24  error    JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  38:16  error    JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
  39:16  error    JSX attribute values should not contain objects created in the same scope  react-perf/jsx-no-new-object-as-prop
```

anajavi:
By "errors out" I meant those errors generated by this plugin.

I believe ChartProvider and BaseChart errors are already fixed in #216.

Some of those eslint errors can be fixed with https://github.com/alexreardon/memoize-one, maybe event the problematic Contexts.

anajavi:
I think it is better to fix these problems in #216 instead of here. It is much cleaner to fix these when doing changes to context providers.

I would like to turn these into warnings and merge this to master. After all of these are fixed,  I will turn these back to errors. Is it okay?

anajavi:
I managed to fix all the problems reported by eslint-react-perf by using memoize-one. This way there should not be breaking changes.

@whawker what do you think?

whawker:
Looks good to me, nice work!!! Just pulling the changes to test it locally :+1:

whawker:
Yep, perfect as far as I can see.

My only suggestion would be changing

```js
const createGetChart = memoizeOne((chart, chartType) => {
    return () => ({
```

to
```js
const createGetChart = memoizeOne((chart, chartType) => () => ({
```

anajavi:
I fixed the code style.

Prettier might be able to format this kind of code style automatically. Needs a reasonable config though.

Merging this now.

PR #224, opened 2019-08-05
Add back missing @babel/runtime

anajavi:
Babel runtime is missing from dependencies. I guess it was removed accidentally by https://github.com/whawker/react-jsx-highcharts/commit/fcb445ff0577341c2d76206682c054d642f5a6bd

This PR adds it back.

Probably needs a release, otherwise some people might have problems using the library.


whawker:
:man_facepalming: don't know how I managed that! Good spot!

PR #227, opened 2019-09-10
Fix Tooltip update not working in Highcharts 7

### LINKED ISSUE #226: Unable to toggle tooltip visibility
#### Description
There seems to be an issue with toggling tooltip visibility using the 'enabled' prop. I tried three separate options.
- Option 1: Pass tooltip as a prop to <HighchartsChart> wrapper directly. This does not work since componentDidUpdate in BaseChart.js only checks if 'plotOptions' changes, it does not check for the 'tooltip' prop.
- Option 2: Conditionally render <Tooltip> component. Not sure why this does not work.
- Option 3: Render <Tooltip> with the 'enabled' prop. Not sure why this does not work either.
In my opinion all three options should work.

https://stackblitz.com/edit/base-chart-example-8k117q?file=MyChart.js
(please uncomment/comment out the code as appropriate to reproduce with each option)

#### Versions (latest at current time)
* React JSX Highcharts version: 3.6.1
* Highcharts version: 7.2.0
* React version: 16.9

#### Workarounds found
- Interestingly option 3 works with Highcharts v6.2.0, something changed with v7? But I want to use the latest version of Highcharts.
- Also, if (~option 1 and option 2~) or (option 1 and option 3) are present it seems to work, but I don't know if this is intended.
- Of course there are other methods, e.g. setting classnames to hide DOM elements, etc but they seem hacky.

anajavi:
Fixes #226

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
+++ b/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
@@ -43,10 +43,8 @@ class Tooltip extends Component {
 
   updateTooltip = config => {
     const chart = this.props.getChart();
-    chart.update({
-      tooltip: config
-    }, false);
-    this.props.needsRedraw();
+    const tooltip = chart.object.tooltip;
+    tooltip.update(config);
   }
 
   render () {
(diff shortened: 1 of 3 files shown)


anajavi:
This is already fixed in v4-alpha. Closing.

PR #234, opened 2019-09-29
Use JavaScript Standard Style

whawker:
No functionality changes, just a code cleanup.

Does introduce `npm run lint-fix` to automatically fix lint issues (where possible)


anajavi:
It's a good idea to have formatter.

I've used [prettier](https://prettier.io) with the following config:
```json
{
  "singleQuote": true,
  "jsxBracketSameLine": true,
  "trailingComma": "es5"
}
```

That config is adapted from [react's repository](https://github.com/facebook/react/blob/master/.prettierrc.js)

Prettier seems to be more used nowadays, probably because it integrates quite well with IDE's.

Both approaches are fine by me. It's easier to write code when you don't have to think about each projects style.

whawker:
Yep, I'm happy with either. Prettier is completely separate from eslint right? So we would be able to keep our React perf plugins etc?

anajavi:
Prettier can be run with or without eslint. [Here's the plugin](https://github.com/prettier/prettier-eslint).

And here's how you run it without eslint:
https://github.com/reduxjs/react-redux/blob/6c873c7e5ab7504ef3d17e68a0cebdde168d8292/package.json#L31

It also can be set up as formatter at least in VSCode and Atom.

anajavi:
Shall we use prettier or eslint-plugin-standard?

Prettier does not offer much of configuration, which can be both bad and good. On the other hand it seems to be quite standard on react projects so most people should have it enabled on IDEs.

Both are really fine by me.

whawker:
I'll open a new PR with Prettier and close this one.

PR #236, opened 2019-09-29
Add WithHighcharts component

anajavi:
Adds WithHighcharts component that could replace last HOC withHighcharts.

usage:
```jsx
import Highcharts from 'Highcharts';

<WithHighcharts Highcharts={Highcharts}>
  <HighchartsChart .... />
</WithHighcharts>
```

### DIFF
--- a/packages/react-jsx-highcharts/src/components/WithHighcharts/index.js
+++ b/packages/react-jsx-highcharts/src/components/WithHighcharts/index.js
@@ -9,9 +9,15 @@ export default function withHighcharts(Component, Highcharts) {
     // ... and renders the wrapped component with the context theme!
     // Notice that we pass through any additional props as well
     return (
-      <HighchartsContext.Provider value={Highcharts}>
+      <HighchartsProvider Highcharts={Highcharts}>
         <Component {...props} />
-      </HighchartsContext.Provider>
+      </HighchartsProvider>
     );
   };
 }
+
+export const HighchartsProvider = ({ Highcharts, children }) => (
+  <HighchartsContext.Provider value={Highcharts}>
+    {children}
+  </HighchartsContext.Provider>
+);
--- a/packages/react-jsx-highcharts/src/index.js
+++ b/packages/react-jsx-highcharts/src/index.js
@@ -1,6 +1,9 @@
 import PropTypes from 'prop-types';
 import withSeriesType from './components/WithSeriesType';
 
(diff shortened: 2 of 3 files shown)


packages/react-jsx-highcharts/src/components/WithHighcharts/index.js
```diff
@@ -16,3 +16,9 @@ export default function withHighcharts(Component, Highcharts) {
     );
   }
 }
+
+export const WithHighcharts = ({ Highcharts, children }) => (
```
whawker: Should we change the default export above to render

```jsx
export default function withHighcharts(Component, Highcharts) {
  // Maybe add a deprecate notice here? Although I feel this HOC component still has some value
  return props => (
      <WithHighcharts Highcharts={Highcharts}>
        <Component {...props} />
      </WithHighcharts>
    )
  }
}
```

anajavi: Done.

anajavi:
I am not quite fan of naming the component WithHighcharts, but can't think of a better name.

The Highcharts prop could also be passed to HighchartsChart component.

whawker:
How about `HighchartsProvider` as an alternative name, similar to Redux? Or maybe `InjectHighcharts`?

> The Highcharts prop could also be passed to HighchartsChart component.

I've been thinking about this one too, and I can think of a few advantages
* We could provide a default value
* Less components in the tree

One thing I do like about the `withHighcharts` HOC or `HighchartsProvider` added here, is that you can add it near the root of your React application, and you never have to think about it again (sort of, adding extra Highcharts modules can trip you up).

I'm trying to weigh up the support burden, if we allowed both...

anajavi:
> One thing I do like about the `withHighcharts` HOC or `HighchartsProvider` added here, is that you can add it near the root of your React application, and you never have to think about it again (sort of, adding extra Highcharts modules can trip you up).
> 
> I'm trying to weigh up the support burden, if we allowed both...

Allowing both would probably cause confusion and hard to debug bugs.. Maybe the HOC and HighchartsProvider are better in the end.

anajavi:
Renamed to HighchartsProvider and added HighmapsProvider to complement it.

anajavi:
@whawker can this be merged?

whawker:
Yep, all good!

Issue #228, opened 2019-09-20
React JSX Highcharts v4 Alpha Feedback (Hooks rewrite)


whawker:
This issue exists to feedback any issues discovered while testing the new version 4 alphas.

Please note, this rewrite is based on [React Hooks](https://reactjs.org/docs/hooks-intro.html), so requires React 16.8 as minimum. Please see the [release notes](https://github.com/whawker/react-jsx-highcharts/releases/tag/v4.0.0-alpha.1) for more information on breaking changes.

leonardolessa:
I had an issue trying to updating chart series while rendering a PlotLine:

`Uncaught TypeError: axis.removePlotLine is not a function
    at usePlotBandLineLifecycle.js:63
    at commitHookEffectList (react-dom.development.js:21073)
    at commitPassiveHookEffects (react-dom.development.js:21106)
    at HTMLUnknownElement.callCallback (react-dom.development.js:363)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:412)
    at invokeGuardedCallback (react-dom.development.js:466)
    at flushPassiveEffectsImpl (react-dom.development.js:24196)
    at unstable_runWithPriority (scheduler.development.js:674)
    at runWithPriority$2 (react-dom.development.js:11834)
    at flushPassiveEffects (react-dom.development.js:24167)
    at react-dom.development.js:23728
    at scheduler_flushTaskAtPriority_Normal (scheduler.development.js:450)
    at flushTask (scheduler.development.js:503)
    at flushWork (scheduler.development.js:635)
    at performWorkUntilDeadline (scheduler.development.js:237)
    at onTimeout (scheduler.development.js:289)`

I'll try to make an stackblitz of it as soon as I can but right now I'm a little bit busy, I hope this can be useful.

anajavi:
@leonardolessa don't bother with the stackblitz. I already found the bug. It should be axis.removePlotBandOrLine not removePlotLine.

Thank you very much for trying out the alpha!

anajavi:
@whawker Can you make a new alpha release when you have time? There's a couple of fixes and optimizations

whawker:
Yep of course. 

FYI The funeral was yesterday, so I should be more available to assist from now on.

anajavi:
Sorry to hear about your loss. Take your time.

whawker:
Thanks @anajavi 

Published:

* `react-jsx-highcharts@4.0.0-alpha.2`
[...]

PR #237, opened 2019-09-29
Export usePlotBandLine to be used as extension point

anajavi:
This allows library users to extend plotline within their own components.

As discussed in https://github.com/whawker/react-jsx-highcharts/issues/228#issuecomment-536346559


packages/react-jsx-highcharts/src/components/PlotBandLine/index.js
```diff
@@ -1,6 +1,7 @@
 import PlotBand from './PlotBand';
 import PlotLine from './PlotLine';
 import PlotBandLineLabel from './PlotBandLineLabel';
+import usePlotBandLine from './usePlotBandLine';
 
 const ChartPlotBand = PlotBand;
```
whawker: I think we could drop the `Chart` prefixes on these variables in the index files - they made sense when they were the output of a `provideChart` HOC, but probably just a weird legacy artefact now.

packages/react-jsx-highcharts/src/components/PlotBandLine/index.js
```diff
@@ -1,6 +1,7 @@
 import PlotBand from './PlotBand';
 import PlotLine from './PlotLine';
 import PlotBandLineLabel from './PlotBandLineLabel';
+import usePlotBandLine from './usePlotBandLine';
```
whawker: As already stated by Travis, looks like this file hasn't been added?

whawker:
Should the `usePlotBandLineLifecycle.js` file begin with a capital 'U' for consistency?

packages/react-jsx-highcharts/src/components/PlotBandLine/index.js
anajavi: Sorry for that. Added

anajavi:
> Should the `usePlotBandLineLifecycle.js` file begin with a capital 'U' for consistency?

I can rename all these. I just named it after a function name, as react kind of forces hook functions to start with uncapitalized 'use'.

During the rewrite I also thought about flattening the src directory structure, because most of the components don't need index.js anymore. Decided to suggest that later, as now it is still easy to see changes from classes to hooks on git.

packages/react-jsx-highcharts/src/components/PlotBandLine/index.js
anajavi: Done.

anajavi: It turned these consts are actually needed. Otherwise they were dropped from production build. I guess it has something to do with webpack tree shaking.

I reverted the change.

PR #238, opened 2019-09-30
Use Prettier code styling

whawker:
Supersedes #234


anajavi:
I'll try to review this soon. Travis seems to be happy with linting, but on my computer it shows:
```
✖ 439 problems (432 errors, 7 warnings)
  432 errors and 0 warnings potentially fixable with the `--fix` option.
```

I also like single quotes more than double quotes, but that is just matter of taste. I can live with both styles :)

I think that IDEs are automatically using prettier when prettierrc is found in project, I will check that one out.

whawker:
Hadn't even spotted `.prettierrc`, I prefer single quotes too. I assumed as it was opinionated I didn't have a choice.

Will update.

anajavi:
Yeah. The project needs to have .prettierrc in root to stop prettier finding prettierrc from parent directory. That was causing the error on my computer.

https://prettier.io/docs/en/configuration.html

anajavi:
That looks nice to me.

Prettier can also format markdown, and code within code blocks in README.md. Maybe that should be done too?

I suggest that you squash & merge this PR in github, otherwise the git history is going to be mess with two code formats.

whawker:
Yep, planning to rebase on master and squash the commits. I should probably reformat the examples too.

PR #240, opened 2019-10-01
Small clean up

whawker:


packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
```diff
@@ -1,55 +1,32 @@
...
   const [rendered, setRendered] = useState(false);
   const domNodeRef = useRef(null);
   const chartRef = useRef(null);
-  const providerValueRef = useRef(null);
+  const providedChartRef = useRef(null);
 
   useLayoutEffect(() => {
     const myChart = initHighcharts(restProps, domNodeRef.current);
     chartRef.current = myChart;
+    providedChartRef.current = createProvidedChart(
```
whawker: Move into `createProvidedChart` for consistency

packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -28,11 +28,11 @@ const Series = memo(
...
-  if (defaultTo(restProps.requiresAxis, true)) {
-    const axis = getAxis();
-    if(!axis) throw new Error(`Series type="${restProps.type}" should be wrapped inside Axis`);
-  }
-*/
+      if (defaultTo(restProps.requiresAxis, true)) {
+        const axis = getAxis();
+        if(!axis) throw new Error(`Series type="${restProps.type}" should be wrapped inside Axis`);
+      }
+    */
```
whawker: Looks like this should be added back?

packages/react-jsx-highcharts/src/components/UseSeries/index.js
```diff
@@ -1,16 +1,15 @@
...
 
 export default function useSeries(seriesId) {
-  const providedSeries = useContext(SeriesContext);
-
+  const contextSeries = useContext(SeriesContext);
   const chart = useChart();
 
   const createStateSeries = () => {
-    if (providedSeries) return createProvidedSeries(providedSeries);
+    if (contextSeries) return createProvidedSeries(contextSeries);
```
whawker: This differs from `UseAxis`, [the create function isn't called](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/src/components/UseAxis/index.js#L12). Is this intentional?

packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
anajavi: That's better.

packages/react-jsx-highcharts/src/components/Series/Series.js
anajavi: It is disabled because useAxis can actually return null. That is when the axis has yet to be created.

I could not figure how to check that no axis will ever be created. Maybe rewriting whole of axis creation to use events, but that would be quite a task.

Some kind of error would be really nice to help debugging.

packages/react-jsx-highcharts/src/components/UseSeries/index.js
anajavi: Nice find! It is definitely a bug.

anajavi: The useAxis is correct.

whawker:
Just fixed the useSeries bug, and changed it so Prettier also formats the tests.

PR #241, opened 2019-10-11
discussion: Add some example docs for docusaurus

anajavi:
This PR is just for discussion purposes in order to have directory structure and conventions for the component docs.

I just quickly added some component docs just to see how the structure feels to use.

I did not see need for separate docs and api, so I just squashed them together.

### DIFF
--- a/docs/api/caption.md
+++ b/docs/api/caption.md
@@ -1,0 +1,19 @@
+---
+id: caption
+title: Caption
+sidebar_label: Caption
+---
+
+The chart's caption, which will render below the chart.
+
+[ Highcharts caption reference](https://api.highcharts.com/highcharts/caption)
+
+## Usage
+
+```jsx
+<HighchartsChart>
+  <Caption align="center">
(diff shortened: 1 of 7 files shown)


anajavi:
Found something interesting, which is included in the coming [docusaurus v2](https://v2.docusaurus.io/docs/markdown-features#embedding-react-components).

React components can be embedded in page using [MDX](https://mdxjs.com/). That would be really helpful when writing component api docs. No need to embed stackblitz examples then.

whawker:
Docs structure looks good, although I wonder if we should completely split the 3 variants to avoid confusion and keep things concise?

That MDX stuff looks cool - I saw the docusaurus v2 note, but didn't thing to check it out. It looks pretty close, I'm think we should skip v1 entirely to avoid future work.

anajavi:
The work-in-progress react-leaflet docusaurus site includes live editor: https://react-leaflet-v3.now.sh

PR #250, opened 2019-11-04
Allow Chart event handlers to be updated

anajavi:
This allows chart event handlers to be updated.

When parent component is hooks based, the passed in event handler might need to change:
```javascript
const onSelection = useCallback(() => {
  // do something with prop that changes
}, [props.somechangingprop]);
```

### DIFF
--- a/packages/react-jsx-highcharts/src/components/UseManualEventHandlers/index.js
+++ b/packages/react-jsx-highcharts/src/components/UseManualEventHandlers/index.js
@@ -14,15 +14,17 @@ const useManualEventHandlers = function(props, target) {
   );
 
   if (modifiedEvenHandlers !== false) {
-    Object.entries(modifiedEvenHandlers).forEach(([eventName, newHandler]) => {
+    Object.keys(modifiedEvenHandlers).forEach(eventName => {
       if (previousEventHandlers) {
         const oldHandler = previousEventHandlers[eventName];
         if (oldHandler) {
           Highcharts.removeEvent(target, eventName, oldHandler);
         }
       }
-
-      Highcharts.addEvent(target, eventName, newHandler);
+      const newHandler = modifiedEvenHandlers[eventName];
+      if (newHandler) {
+        Highcharts.addEvent(target, eventName, newHandler);
+      }
     });
   }
 };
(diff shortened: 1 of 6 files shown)


whawker:
The code looks good to me, but could you just give me a bit of background/example where the current set up causes issues?

anajavi:
The previous setup did not update event handlers to highcharts.

Event handlers don't change when parent is class based component. That is, the eventHandler never changes:
```javascript
class Component {
  eventHandler = (event) => {
    this.props.eventHandler(event, true);
  }
}
```

But with hooks the eventHandler can be actually changed because some dependency changes:
```javascript
const eventHandler = useCallback((event) => {
  props.eventHandler(event, true);

},
[props.parentEventHandler]); // useCallback creates new function when props.parentEventHandler changes
```

That's why it is way easier to pass the new eventhandler to highcharts. Otherwise the parent component needs to make sure that the eventhandler never changes, and it can be hairy sometimes.

I've been actually thinking, that I could make a custom hook that updates the event handlers:
```javascript
function useModifiedEvents(props) {
  // filter props for changed event handlers
  // call addEventHandlersManually
}
```
That hook could then be used in axis, series etc to update eventhandlers.

anajavi:
I have to rewrite this so, that it removes only event handlers added by user. Otherwise it might break highcharts inner workings.

packages/react-jsx-highcharts/src/components/UseManualEventHandlers/index.js
```diff
@@ -0,0 +1,30 @@
...
+  const eventHandlers = getEventsConfig(props);
+  const previousEventHandlers = usePrevious(eventHandlers);
+
+  const modifiedEvenHandlers = getModifiedProps(
+    previousEventHandlers,
+    eventHandlers
+  );
+
+  if (modifiedEvenHandlers !== false) {
+    Object.entries(modifiedEvenHandlers).forEach(([eventName, newHandler]) => {
```
whawker: Think we should use Lodash [`toPairs`](https://lodash.com/docs/4.17.15#toPairs) as Edge 12 & 13 doesn't support this.

anajavi: v.4 was requiring polyfills anyway, but I can rewrite that with Object.keys too.

PR #251, opened 2019-11-21
Fix navigator

anajavi:
This fixes navigator crash when updated (https://github.com/whawker/react-jsx-highcharts/issues/228#issuecomment-551308361).

The root cause was that navigator re-creates it's axes when navigator is updated.

There were several problems:
1. 3e65a8d creates axis earlier, so that navigator can access the parent series axis
2. b8e74fd allows provided axis to change, so that plotbands can react to axis changes
3. cf538e7 forces plotbands to re-render when axis was changed
4. e1d3ff1 fixes the navigator itself

Related issue on highcharts repository: https://github.com/highcharts/highcharts/issues/12406

Probably more axis consumers, like Annotations and maybe Series need updating, but maybe we should test this in another alpha first?

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Axis/Axis.js
+++ b/packages/react-jsx-highcharts/src/components/Axis/Axis.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import uuid from 'uuid/v4';
 import { attempt } from 'lodash-es';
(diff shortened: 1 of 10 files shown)


whawker:
If we rebase this branch on master, I'll publish an alpha from it, then after testing we decide whether to merge or not?

anajavi:
> If we rebase this branch on master, I'll publish an alpha from it, then after testing we decide whether to merge or not?

I rebased this. Any chance we can include #250 in the alpha too or just merge it to master first.

whawker:
Just this second merged #250, any chance you could rebase again? (Sorry!)

I'll publish one alpha from master, and one alpha from this branch.

anajavi:
> Just this second merged #250, any chance you could rebase again? (Sorry!)
> 
> I'll publish one alpha from master, and one alpha from this branch.

Rebased. And great!

whawker:
Published master as
* `react-jsx-highcharts@4.0.0-alpha.8`
* `react-jsx-highstock@4.0.0-alpha.8`
* `react-jsx-highmaps@2.0.0-alpha.8`

And this branch as (note **dev**)
* `react-jsx-highcharts@4.0.0-dev.1`
* `react-jsx-highstock@4.0.0-dev.1`
* `react-jsx-highmaps@2.0.0-dev.1`

anajavi:
react-jsx-highcharts@4.0.0-dev.1 crashes LineSeries, but only sometimes. I'll try to make a reproduction of it.

anajavi:
This won't work.

Under some conditions react runs the render twice on mount. On second render the axisRef.current does not keep the previous renders value. This leads to creating the axis on both renders.

So this PR needs to be scrapped. Also the ColorAxis needs to be fixed to use useEffect too.

I am out of ideas how to fix the Navigator now.

PR #253, opened 2019-12-03
Create coloraxis inside useEffect

anajavi:
As found in #251, creating highcharts objects outside of useEffect is not safe in all cases.

Quite straightforward, I think it can be merged immediately.


whawker:
Nice work!

PR #255, opened 2019-12-05
Replace useDelayOnce with simple useEffect

anajavi:
This removes useDelayOnce. Now the code depends on React doing useEffect calls in the order the Components are declared in JSX.

After this it is not possible to reference axis or series before they are rendered in document:
```jsx
<LineSeries axisId="myAxis" /><!-- does not work -->
<YAxis id="myAxis"/>
```

Now the code does not depend on timing of renders, which can change when concurrent mode lands in react.

Also this adds the same logic to coloraxis.


anajavi:
This is the last breaking change I have in my mind. I hope after this we can wrap things up and release final version?

whawker:
Looks good to me! Sorry for the delay!

whawker:
@anajavi I think we're ready to release too, I will do it tomorrow though (16th) if thats ok.

anajavi:
@whawker great! In the end it is the only way to uncover the real bugs :)

PR #257, opened 2019-12-11
Drop support for highchart < 8

anajavi:
Highcharts 8 doesn't seem to have any breaking changes, so this isn't technically needed now.

On the other hand Highcharts 7 doesn't get updates anymore, so it would be easier to drop support now if anything in Highchart 8 helps us to make simpler code.

What do you think?


whawker:
Yep, agreed we should drop support for 7, gives us a nice clean break for our new major version

PR #260, opened 2019-12-17
Fix PlotBand.Label updating together with axis

anajavi:
When both Axis and and PlotBand.Label are updated **twice**, the label crashes.

The root cause: Updating axis recreates plotbands and -lines. Also calling axis.update without redrawing chart leaves axis.plotBandAndLines array empty.

**This PR makes axis updates redraw synchronously when axis has plotbands or plotlines. Also this changes usePlotBandLine to always return newest object instance.**

Note: PlotBand label can be updated once. After that it does not render changes:
https://jsfiddle.net/8tdozmgf/


packages/react-jsx-highcharts/src/components/Axis/Axis.js
```diff
@@ -36,8 +36,14 @@ const Axis = ({ children = null, dynamicAxis = true, ...restProps }) => {
...
-      axis.update(modifiedProps, false);
-      chart.needsRedraw();
+      // if there are plotlines or bands, the chart needs to be redrawn before
+      // they can be accessed
+      if (axis.plotLinesAndBands && axis.plotLinesAndBands.length > 0) {
+        axis.update(modifiedProps, true);
+      } else {
+        axis.update(modifiedProps, false);
+        chart.needsRedraw();
+      }
```
anajavi: After axis.update call and before chart.redraw call, the axis.plotLinesAndBands is empty.

this is hacky, but I see no other way..

packages/react-jsx-highcharts/src/components/PlotBandLine/UsePlotBandLineLifecycle.js
```diff
@@ -25,8 +25,15 @@ export default function usePlotBandLineLifecycle(props, plotType) {
...
+      axis.addPlotBandOrLine(opts, plotType);
       setPlotbandline({
-        object: axis.addPlotBandOrLine(opts, plotType)
+        id: myId,
+        getPlotBandLine: () => {
+          const plotbandlineObject = axis.object.plotLinesAndBands.find(
+            plb => plb.id === myId
+          );
+          return plotbandlineObject;
+        }
```
anajavi: This is technically a breaking change, but then again it did not work at all before.

packages/react-jsx-highcharts/src/components/PlotBandLine/UsePlotBandLineLifecycle.js
```diff
@@ -25,8 +25,17 @@ export default function usePlotBandLineLifecycle(props, plotType) {
...
-        object: axis.addPlotBandOrLine(opts, plotType)
+        id: myId,
+        getPlotBandLine: () => {
+          if (axis && axis.object && axis.object.plotLinesAndBands) {
+            const plotbandlineObject = axis.object.plotLinesAndBands.find(
+              plb => plb.id === myId
+            );
+            return plotbandlineObject;
+          }
+        }
```
anajavi: This is technically a breaking change as it changes the object returned by usePlotBandLine.

 But then again it did not work at all before.

anajavi: Adding object getter might actually work  here instead.

packages/react-jsx-highcharts/src/components/PlotBandLine/UsePlotBandLineLifecycle.js
```diff
@@ -25,8 +25,18 @@ export default function usePlotBandLineLifecycle(props, plotType) {
...
+        id: myId,
+        get object() {
+          if (axis && axis.object && axis.object.plotLinesAndBands) {
+            const plotbandlineObject = axis.object.plotLinesAndBands.find(
+              plb => plb.id === myId
+            );
+            return plotbandlineObject;
+          }
+          return null;
+        }
```
anajavi: This is needed because updating parent axis recreates all plotbands and lines.

packages/react-jsx-highcharts/test/components/PlotBandLine/PlotBand.integration.spec.js
```diff
@@ -0,0 +1,174 @@
...
+  const AxisSpy = ({ axisId, axisRef }) => {
+    const axis = useAxis(axisId);
+    const addPlotBandOrLineSpy = jest.spyOn(axis, 'addPlotBandOrLine');
+    const removePlotBandOrLineSpy = jest.spyOn(axis, 'removePlotBandOrLine');
+
+    axisRef.current = axis;
+    axisRef.addPlotBandOrLineSpy = addPlotBandOrLineSpy;
+    axisRef.removePlotBandOrLineSpy = removePlotBandOrLineSpy;
+    return null;
+  };
```
anajavi: In the future this could be extracted into a separate component along with ChartSpy and SeriesSpy. 

Useful for writing integration tests.

packages/react-jsx-highcharts/src/components/PlotBandLine/UsePlotBandLineLifecycle.js
whawker: Could we add this as a comment in the code to make that clear?

packages/react-jsx-highcharts/src/components/PlotBandLine/UsePlotBandLineLifecycle.js
```diff
@@ -25,8 +25,18 @@ export default function usePlotBandLineLifecycle(props, plotType) {
...
+      axis.addPlotBandOrLine(opts, plotType);
       setPlotbandline({
-        object: axis.addPlotBandOrLine(opts, plotType)
+        id: myId,
+        get object() {
+          if (axis && axis.object && axis.object.plotLinesAndBands) {
+            const plotbandlineObject = axis.object.plotLinesAndBands.find(
+              plb => plb.id === myId
+            );
+            return plotbandlineObject;
```
whawker: Could return on line 33 rather than assigned the result to a const?

packages/react-jsx-highcharts/test/components/PlotBandLine/PlotBand.integration.spec.js
whawker: Probably worth doing now? Seems like something that will end up being forgotten otherwise

anajavi:
Fixed all the review comments. I consider this ready for merge.

PR #261, opened 2019-12-23
Fix importing in SSR app

anajavi:
This should fix the issue reported in https://github.com/whawker/react-jsx-highcharts-examples/issues/2

Importing debounce-raf caused exception when window was not available.


mg-code:
@whawker when 4.0.1 is going to be released?

We upgraded package to latest version and now our build breaks.

PR #280, opened 2020-04-23
Upgrade to prettier v2

anajavi:
Upgrade to prettier v2. 

Prettier v2 changed some formatting defaults which I mostly configured back.

Another change is [default endOfLine](https://prettier.io/docs/en/options.html#end-of-line). It can be fixed with either configuring it to auto, or:
> Add * text=auto eol=lf to the repo's .gitattributes file. You may need to ask Windows users to re-clone your repo after this change to ensure git has not converted LF to CRLF on checkout.

Either one is still needed for this PR, which one you prefer?

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,6 @@
 {
   "singleQuote": true,
-  "tabWidth": 2
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "arrowParens": "avoid"
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "singleQuote": true,
   "tabWidth": 2,
   "trailingComma": "none",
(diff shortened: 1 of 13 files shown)


whawker:
I develop on a Mac personally so I'm happy with either. But I guess the first option is the smallest change to what we have in our gitattributes currently? So I'd go with that

anajavi:
> I develop on a Mac personally so I'm happy with either. But I guess the first option is the smallest change to what we have in our gitattributes currently? So I'd go with that

done. merge?

anajavi:
> Looks good! Although do we not need to update the gitattributes file itself?

Update to gitattributes is not needed when prettier is set to auto. Generally the gitattributes is kind of better way, but changing it causes havoc.

I will merge this now. Thanks for the late night review!

PR #281, opened 2020-04-27
Expose 'updatePoints' parameter of setData

veloek:
To allow overriding the default value (true) when calling setData.

In our project, with animations turned off, we don't need to update points for any visual effect and we noticed a huge performance improvement on large data sets with `updatePoints` set to false.

Please let me know if there are other changes throughout the codebase necessary as well and feel free to push updates on this branch.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -25,6 +25,7 @@ const Series = memo(
     children = null,
     axisId,
     requiresAxis = true,
+    updatePoints = true,
     ...restProps
   }) => {
     const seriesProps = { id, data, type, visible, ...restProps };
@@ -80,7 +81,7 @@ const Series = memo(
       let doRedraw = false;
       // Using setData is more performant than update
       if (isDataEqual(data, prevProps.data) === false) {
-        series.setData(data, false);
+        series.setData(data, false, undefined, updatePoints);
         doRedraw = true;
       }
       if (visible !== prevProps.visible) {
@@ -118,7 +119,8 @@ Series.propTypes = {
   visible: PropTypes.bool,
   children: PropTypes.node,
   axisId: PropTypes.string,
-  requiresAxis: PropTypes.bool
+  requiresAxis: PropTypes.bool,
+  updatePoints: PropTypes.bool
 };
 
 const getSeriesConfig = (props, axis, colorAxis, requiresAxis) => {
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -25,7 +25,7 @@ const Series = memo(
     children = null,
     axisId,
     requiresAxis = true,
-    updatePoints = true,
+    updatePoints,
     ...restProps
   }) => {
     const seriesProps = { id, data, type, visible, ...restProps };
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -25,7 +25,7 @@ const Series = memo(
     children = null,
     axisId,
     requiresAxis = true,
-    updatePoints,
+    jsxOptions,
     ...restProps
   }) => {
     const seriesProps = { id, data, type, visible, ...restProps };
@@ -81,7 +81,9 @@ const Series = memo(
       let doRedraw = false;
       // Using setData is more performant than update
       if (isDataEqual(data, prevProps.data) === false) {
-        series.setData(data, false, undefined, updatePoints);
+        const animation = jsxOptions && jsxOptions.animation;
+        const updatePoints = jsxOptions && jsxOptions.updatePoints;
+        series.setData(data, false, animation, updatePoints);
         doRedraw = true;
       }
       if (visible !== prevProps.visible) {
@@ -120,7 +122,10 @@ Series.propTypes = {
   children: PropTypes.node,
   axisId: PropTypes.string,
   requiresAxis: PropTypes.bool,
-  updatePoints: PropTypes.bool
+  jsxOptions: PropTypes.shape({
+    animation: PropTypes.bool,
+    updatePoints: PropTypes.bool
+  })
 };
 
 const getSeriesConfig = (props, axis, colorAxis, requiresAxis) => {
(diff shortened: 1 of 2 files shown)


packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -25,6 +25,7 @@ const Series = memo(
     children = null,
     axisId,
     requiresAxis = true,
+    updatePoints,
```
whawker, line 28: Looks like the default for this is `true`, so I think the props should reflect that

https://api.highcharts.com/class-reference/Highcharts.Series#setData

packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -80,7 +81,7 @@ const Series = memo(
       let doRedraw = false;
       // Using setData is more performant than update
       if (isDataEqual(data, prevProps.data) === false) {
-        series.setData(data, false);
+        series.setData(data, false, undefined, updatePoints);
```
whawker, line 84: Think we should support the third arg (animation) here via a prop too? Which would default to `undefined`

anajavi: [The "animation" can be passed to series as option.](https://api.highcharts.com/highcharts/series.line.animation)

So different name would be needed.

anajavi:
> Thanks for the contribution @veloek! Added a few thoughts.
> 
> Any comments @anajavi?

I like this! Nice work @veloek.

Not sure about naming of updatePoints prop. To prevent name collisions, maybe it should be passed within some kind of options object. [At least the animation prop would collide with Series props.](https://api.highcharts.com/highcharts/series.line.animation)

I'll see if I find time to try this with some big datasets.

packages/react-jsx-highcharts/src/components/Series/Series.js
whawker: Good spot!

whawker: Maybe `animated` as a prop name?

packages/react-jsx-highcharts/src/components/Series/Series.js
veloek: Yeah, I was a bit back and forth regarding this. On one side it makes sense to have the default the  same as Highcharts, but on the other hand I thought passing undefined would be more future proof, as Highcharts could decide to flip the default (though it seems pretty unlikely) and this library should just wrap Highcharts, not implement it's own defaults. I'm fine with it either way, you decide 🙂

anajavi:
> I'll see if I find time to try this with some big datasets.

I did not find anything suitable to test this with. But I think it's fine, provided that the naming can be decided on.

Maybe something like this? The name here is arbitrary as I cannot think of a proper one.
```jsx
const jsxOptions = { updatePoints: false };
<Series jsxOptions={jsxOptions} />
```

veloek:
Yeah, I could try wrapping it in an object - that would make room for the animation option as well 🙂 

I don't have any strong opinions with regards to naming. One could think to call it `setDataOptions`, but that would make it very specific and any future options would need their own "options" prop, so I think `jsxOptions` is fine. Then the same name could be used in other components if needed and it'd be consistent and familiar.

anajavi:
@whawker can you think of a good name for the options object?

whawker:
I would prefer to avoid an object if possible, but that's just a personal preference. This whole project started because of my dislike for config objects 😂

Would it be too verbose to have `dataAnimation` and `dataUpdatePoints` as props? I guess they're not likely to change at runtime though, plus we're deviating from Highcharts API which I like to avoid.

If an object is preferable here, I would suggest `dataOptions` as a name.

veloek:
@anajavi, @whawker, are we any closer to a decision on this? Just let me know which way you want to go 🙂

anajavi:
> @anajavi, @whawker, are we any closer to a decision on this? Just let me know which way you want to go 🙂

Sorry for the delay @veloek, I've been swamped with other things lately. Here the Rome should be built no later than midsummer - every year.

I saw the jsxOptions as generic name for react-jsx-highcharts specific configuration. That name could be used in other components too, and it would be clear that it is not something that is passed to highcharts.

But then again, both ways are fine by me. @whawker can you say a final word for this, so that this can be wrapped up?

whawker:
@anajavi makes a good point, and certainly would be easier to document!

Let's go with a `jsxOptions` object.

anajavi:
> Let's go with a `jsxOptions` object.

I guess this is ready to be merged then. @whawker can you make a release? There are quite many changes waiting for a release.

whawker:
Sorry for the delay all, life busy on my end too at the moment. Hopefully will have time next week to give this project some love (docs, examples updates etc).

I'll aim to release 4.1 tomorrow morning (UK time)

whawker:
Just released

* react-jsx-highcharts 4.1.0
* react-jsx-highstock 4.1.0
* react-jsx-highmaps 2.1.0

Writing a release note as we speak.

whawker:
@veloek do you have a small example showing the difference this change makes? So I can add it to the release notes?

whawker:
P.s. just realised the `jsxOptions.animation` property will never work here, due to how we handle redraws internally.

In vanilla Highcharts the `animation` argument is eventually passed through to `chart.redraw(animation)`. 

However in RJH we enforce `redraw = false`, and call `needsRedraw` instead. The animation argument isn't passed through to `needsRedraw` - even if we did, it would make little difference due to redraw being debounced, so redraw might not be called anyway.

I'm not sure there is an easy solution to this.

veloek:
@whawker, sorry I don't have any example to share. I tried to replicate the behavior we see in our solution in a highcharts demo, but couldn't reproduce it there and didn't have too much time to fiddle around.

The `animation` property was never something I was concerned with, so I wouldn't mind if you removed it. It would leave the config object with a single prop though, so perhaps it would make sense to go back to the original version. Too bad this didn't come up before the 4.1.0 release. Not sure how you feel about changing this now.

whawker:
Ah, so is that to say the `updatePoints` part of this PR doesn't work either?

veloek:
@whawker, no the `updatePoints` part is working just fine and solving my issue with `setData` 🙂

whawker:
Release note for reference: https://github.com/whawker/react-jsx-highcharts/releases/tag/v4.1.0

veloek:
Cool to see that you managed to create a simple example to demonstrate the issue, @whawker 🙂 Have you made any attempt to reproduce the same behavior in vanilla Highcharts?

whawker:
I took [this vanilla Highcharts example](https://jsfiddle.net/BlackLabel/v1hga67t/) and converted it

veloek:
Ah, cool. Somehow I missed this issue: highcharts/highcharts#11016

Issue #282, opened 2020-05-04
How to obtain the same responsive of official package


lezan:
#### Bug or Feature Request?
I do not know if it is a bug or intended behaviour, but I am not able to figure out how to obtain the same result.

#### Description
I am switching from react-official to react-jsx because with the last I can have more performance on update, but I am some issue obtain the same responsability.
With the official package I wrap chart into a container with responsability, and then I set  HighchartsReact container with height and width of 100%, something like that.
```
const style = {
    height,
    width: '100%',
};

<Row>
    <Col>
        <div id={nameChart} className="chart" style={style}>
            { chartElement !== null ? chartElement : null }
        </div>
    </Col>
</Row>
```
and then
```
<HighchartsReact
    highcharts={Highcharts}
    constructorType="chart"
    containerProps={{ style: { height: '100%', width: '100%' } }} // <- This one.
    options={chartOptions}
/>
```
With react-jsx I try to the same, I just switch the last part
```
<HighchartsChart
     containerProps={{ style: { height: '100%', width: '100%' } }}
    plotOptions={plotOptions}
>
    <Chart
        type={typeChart}
    />
    ...
```
Doing like that I am not able to obtain responsability on height, because chart does not change his dimension.
I also try to do 
```
<HighchartsChart
     style={{ height: '100%', width: '100%' }}
    plotOptions={plotOptions}
>
    ...
```
but same result.
#### How to reproduce
Just try to resize window on the two examples.

#### Live demo demonstrating bug
With react-jsx: https://stackblitz.com/edit/base-chart-example-dycbqd.
With highcharts-react: https://stackblitz.com/edit/react-q2junp.

#### Versions
* React JSX Highcharts version: 4.0.1
* Highcharts version: 8.0.4
* React version: 16.13.1

Thanks!

whawker:
Hi @lezan, glad to hear you're interested in switching!

I would personally recommend using Highcharts "[styled by CSS](https://whawker.github.io/react-jsx-highcharts/examples/StyleByCSS/index.html)" (see the `styledMode` prop) rather than the config way above

Then it would be a case of adding a CSS rule for the `.highcharts-container` div, to obtain the behaviour you desire.

If this isn't an option for legacy reasons, I might look into adding `containerProps` support for easier migration from the official package,

lezan:
Hi @whawker, thanks for you answer.
Using styledMode made the trick.

[...]

PR #283, opened 2020-05-06
Add containerProps to BaseChart

lezan:
As discussed [here](https://github.com/whawker/react-jsx-highcharts/issues/282) I add containerProps to BaseChart in order to add style to HighchartsReact and make it responsive relative to parent container.

I tested it and seems is working (with the example reporeted on issue linked before).

Let me know if I need to add more.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
+++ b/packages/react-jsx-highcharts/src/components/BaseChart/BaseChart.js
@@ -11,6 +11,7 @@ const BaseChart = ({
   children = null,
   callback = noop,
   className = '',
+  containerProps = null,
   ...restProps
 }) => {
   const [rendered, setRendered] = useState(false);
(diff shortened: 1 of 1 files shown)


whawker:
P.s. thanks for the contribution @lezan!

anajavi:
> Happy to merge regardless, any thoughts to add @anajavi?

I am happy. The test would serve as api stability guarantee so maybe a really simple test would suffice. But happy to merge regardless too.

lezan:
> 
> 
> It would be good to add a test for this if possible, although it looks like we don't have a test for the `className` prop either.
> 
> Happy to merge regardless, any thoughts to add @anajavi?

I was thinking so (add test for containerProps), but before I checked if you did the same for other props (like className) but having only found [this](https://github.com/whawker/react-jsx-highcharts/blob/master/packages/react-jsx-highcharts/test/components/BaseChart/BaseChart.spec.js). 

I will try to look into jest (are you using it, right? I never) and then try to make a PR with a test case for containerProps.

Thanks for this library, so much good stuff.

whawker:
> I will try to look into jest (are you using it, right? I never) and then try to make a PR with a test case for containerProps.

Jest and Enzyme, yes.

> I will try to look into jest (are you using it, right? I never) and then try to make a PR with a test case for containerProps.

Cool, merging. Look forward to your next PR!

PR #290, opened 2020-05-22
Fix tooltip not updating properly

### LINKED ISSUE #289: Tooltip pointFormat not updating
#### Bug or Feature Request?
Bug

#### Description
Unable to update tooltip format.

#### How to reproduce
1. Open stackblitz below.
2. Attempt to change pointFormat, with the button. The tooltip does not update

https://stackblitz.com/edit/base-chart-example-ko6uke?file=MyChart.js

#### Versions
* React JSX Highcharts version: 4.0.1
* Highcharts version: 8.1.0
* React version: 16.13.1

#### Notes
I believe this code works with an earlier version of highcharts + react-jsx-highcharts.

anajavi:
This fixes #289.

Now the Tooltip does not get recreated on mount, which might lead to settings leaking from previous tooltip options. But now both pointFormat and enabled props work at the same time.

It might be possible to store chart.options.tooltip.userOptions on mount and restore it on unmount, but that might cause more trouble.

@whawker what do you think?

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
+++ b/packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
@@ -1,6 +1,6 @@
 import React, { useEffect, memo } from 'react';
 import PropTypes from 'prop-types';
-import { attempt, defaultTo } from 'lodash-es';
+import { attempt } from 'lodash-es';
 import useChart from '../UseChart';
 import useHighcharts from '../UseHighcharts';
 import useModifiedProps from '../UseModifiedProps';
@@ -11,11 +11,10 @@ const Tooltip = memo(props => {
   const chart = useChart();
   const Highcharts = useHighcharts();
 
-  restProps.enabled = defaultTo(props.enabled, true);
+  restProps.enabled = props.enabled ?? true;
 
   useEffect(() => {
-    const chartObj = chart.object;
-    chartObj.tooltip = new Highcharts.Tooltip(chartObj, {
+    updateTooltip(chart, {
       ...(Highcharts.defaultOptions && Highcharts.defaultOptions.tooltip),
       ...restProps
     });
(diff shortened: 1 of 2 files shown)


packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
```diff
@@ -11,11 +11,10 @@ const Tooltip = memo(props => {
...
   const Highcharts = useHighcharts();
 
-  restProps.enabled = defaultTo(props.enabled, true);
+  restProps.enabled = props.enabled ?? true;
 
   useEffect(() => {
-    const chartObj = chart.object;
-    chartObj.tooltip = new Highcharts.Tooltip(chartObj, {
+    updateTooltip(chart, {
       ...(Highcharts.defaultOptions && Highcharts.defaultOptions.tooltip),
```
whawker, line 18: I know this line isn't part of this PR, but perhaps it should be moved into `updateTooltip` below?

anajavi: I don't think it can be moved there as updateTooltip is called with only modifiedProps too. Then updateTooltip would reset non modifiedprops to defaults.

packages/react-jsx-highcharts/src/components/Tooltip/Tooltip.js
```diff
@@ -36,8 +35,9 @@ const Tooltip = memo(props => {
 });
 
 const updateTooltip = (chart, config) => {
-  const tooltip = chart.object.tooltip;
-  tooltip.update(config);
+  chart.update({
```
whawker, line 38: May need to check that this, doesn't undo https://github.com/whawker/react-jsx-highcharts/issues/226#issuecomment-529985171

whawker: Sorry, just noticed you already referenced this bug on #289 - apologies

PR #294, opened 2020-06-26
Return the map chart instance to the callback prop

rowebrandon:
It looks like the chart that was being returned in the callback prop is the chart that is passed as a prop (thus `undefined` if no chart is passed). This pull request changes the instance returned in the callback and adds a test to ensure that the chart return is defined.


whawker:
As yes, you're right. Thanks for the fix, and the test too!

whawker:
Published `react-jsx-highmaps@2.1.1`

Thanks @rowebrandon!

Issue #299, opened 2020-08-03
Error when using navigator and stacked series


veloek:
#### Bug or Feature Request?
Bug. "Cannot read property 'base' of undefined"

#### Description
Using a navigator with two stacked column series throws an error. It looks like it got introduced with highcharts version 8.1.1 as it works if you revert to 8.1.0.

The bug probably belongs with highcharts, but I didn't manage to reproduce it with vanilla highcharts so I reported it here.

#### How to reproduce
Run the demo.

#### Live demo demonstrating bug
https://stackblitz.com/edit/stockchart-updatepoints-false-qf2hes

#### Versions
* React JSX Highcharts version: 4.1.0
* Highcharts version: 8.1.1
* React version: 16.13.1

veloek:
Here's another example showing an issue that seems to have come with highcharts 8.1.1:
https://stackblitz.com/edit/stockchart-updatepoints-false-1yudy4

For some reason only the series on the first y-axis is rendered. If you revert to highcharts 8.1.0 you can see that it works and both series are there.

Any idea what could've caused these issues, @whawker, @anajavi? Reading the changelog, 8.1.1 seems like a harmless upgrade...

anajavi:
> Here's another example showing an issue that seems to have come with highcharts 8.1.1:
> https://stackblitz.com/edit/stockchart-updatepoints-false-1yudy4

Here's the same problem with vanilla highcharts: https://jsfiddle.net/2mb6vyrn/

The highcharts > 8.1.0 doesn't accept chart.setSize(undefined, undefined) anymore. It is called on this row:
https://github.com/whawker/react-jsx-highcharts/blob/f4d3acbb7e5e3ae2d094149b9ce8febcb706a8df/packages/react-jsx-highcharts/src/components/Chart/Chart.js#L32


Removing ```<Chart />``` from your examples gets rid of the call.

I think this should be fixed in react-jsx-highcharts.

whawker:
Great spot @anajavi! I guess this is also related? https://github.com/highcharts/highcharts/issues/13752

veloek:
I did a bisect of highcharts between 8.1.0 and 8.1.1 and found this to be the first bad commit: https://github.com/highcharts/highcharts/commit/57ce06ee444abb7b9c5e962c68dc6f5f831b64ae

whawker:
[...]

PR #300, opened 2020-08-04
Do not call chart.setSize when both height and width are undefined

anajavi:
This is a workaround for #299.

Also this now allows height and width to be set 0 or null.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Chart/Chart.js
+++ b/packages/react-jsx-highcharts/src/components/Chart/Chart.js
@@ -12,7 +12,7 @@ const Chart = memo(({ type = 'line', width, height, ...restProps }) => {
   const modifiedProps = useModifiedProps({ type, ...restProps });
 
   useEffect(() => {
-    if (!(typeof width === 'undefined' && typeof height === 'undefined')) {
+    if (!(width === undefined && height === undefined)) {
       chart.setSize(width, height);
     }
   }, [width, height]);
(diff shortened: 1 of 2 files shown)


packages/react-jsx-highcharts/src/components/Chart/Chart.js
```diff
@@ -5,31 +5,30 @@ import useModifiedProps from '../UseModifiedProps';
...
 
-const Chart = memo(({ type = 'line', ...restProps }) => {
+const Chart = memo(({ type = 'line', width, height, ...restProps }) => {
   const chart = useChart();
   const mounted = useRef(false);
 
   const modifiedProps = useModifiedProps({ type, ...restProps });
 
+  useEffect(() => {
+    if (!(typeof width === 'undefined' && typeof height === 'undefined')) {
```
whawker, line 15: Just curious why `typeof x === 'undefined'` instead of `x === undefined`?

anajavi: I wasn't thinking. It just came that way :) 

Should I change it?

anajavi: I changed it to ```x === undefined```

anajavi:
@whawker I hope this can be released soon as it is quite annoying bug

whawker:
Published.

* `react-jsx-highcharts@4.1.1`
* `react-jsx-highstock@4.1.2`
* `react-jsx-highmaps@2.1.2`

Issue #267, opened 2020-01-22
Example with react hooks usage?


kirankalyan5:
#### Description

Is there an example of using this library with react hooks?

anajavi:
It depends on what do you want to accomplish using hooks? Can you please describe your use case?

Currently the hooks provided can be used to extend the functionality. For example:
https://whawker.github.io/react-jsx-highcharts/examples/CustomComponent/

kirankalyan5:
Hi, @anajavi thanks for the quick response. I'm currently using `highcharts-react-official` in my project. Now I was planning to introduce more configurable options in terms of having a dynamic chart type, x-axis & y-axis, etc. When I was evaluating I came across your library which looked good(Kudos to you and your team).

In terms of hooks, I saw your recent release notes where you have migrated from HOC to hooks and I was curious to find out more examples as some the current example code were still on HOC's.

PS. I have not yet decided what I will achieve with the help of hooks in this library but I wanted to learn how this library could help achieve the configurable options I'm trying to.

anajavi:
The examples are still using withHighcharts hook, which can be replaced with ```<HighchartsProvider>``` component. The withHighcharts HOC is provided for backwards compatibility.

I don't think you need any hooks for the dynamic chart type and axes. Updating chart type for example was just fixed in v4.0.1 in #265. Axes can be added dynamically too just by adding and removing ```<XAxis>``` and ```<YAxis>``` components.

Here's an example of changing chart type: https://stackblitz.com/edit/react-1zdbcb?file=Hello.js

kirankalyan5:
Thank you, I will have a check.

kirankalyan5:
@anajavi does it have support for typescript, if yes is @types definition a separate package?

I can still `declare module 'react-jsx-highcharts';`  but I lose the benefits of typescript.
I saw #113, is there any outcome?
 Let me know if I have missed anything.

Thanks.

anajavi:
> @anajavi does it have support for typescript, if yes is @types definition a separate package?

There's no typescript typings as far as I know. Highcharts itself seems to have them now, so it might not be such an huge task anymore.

react-jsx-highcharts doesn't have too many typings for itself, most are just passed to highcharts so any typings should just reference highcharts types.

a PR would be welcome.

> I saw #113, is there any outcome?

There are no flow typings for Highcharts, so making them for this library is a huge task.

kirankalyan5:
@anajavi 👍

Issue #113, opened 2018-06-04
Integrate Flow typing


whawker:
Integrate Flow Typing for better IDE support

anajavi:
I added flow tooling in https://github.com/anajavi/react-jsx-highcharts/tree/flow

I tried to make small example of how it would work, and it quickly became extremely complex :)

I have to take some time to think about how to make it simple and easy to use, maybe just bypass most of the HOC typings and leave only typings for the user facing component props.

I don't want to make very complex system out of this as typescript might make flow obsolete.

bondz:
Is there an existing type def for typescript?

whawker:
@bondz not at the moment, I'm not a fan of typescript so I can't find the motivation to add them. Would love a PR though, if you feel up to it?

bondz:
I've written a few type definitions for JS libraries, would give this a go sometime during the weekend and send a PR.

whawker:
@anajavi Nice work!! I also prefer the simpler approach, I would much rather provide flow typing on the external API rather than internally, as it means it's much easier to maintain.

The existing PropType definitions only define the required props for minimal functionality, does it matter if you pass additional props to a component, that aren't referenced in Flow types?

anajavi:
Let's go with the external typings then. I will send a PR when I'm happy with it.

It doesn't seem to matter if you pass additional props, at least in my flow config, but I will check it.

As for props that are passed to highcharts: it is hard to know their types beforehand. Flow typings for highcharts would help, but I'm afraid there is no such thing.

anajavi:
PR in #116

anajavi:
I guess this one has gone stale.

Almost all libraries providing typings are doing it with typescript nowadays. Even Highcharts has typescript typings, which makes typing react-jsx-highcharts with typescript considerably easier.

Closing.

Issue #259, opened 2019-12-16
Version 4 feedback and issues


whawker:
This issue exists to feedback any issues discovered with version 4.

Please note, this rewrite is based on [React Hooks](https://reactjs.org/docs/hooks-intro.html), so requires React 16.8.6  and Highcharts 8 as minimum. Please see the [release notes](https://github.com/whawker/react-jsx-highcharts/releases/tag/v4.0.0) for more information on breaking changes.

filoromz:
Hi great work with the rewrite!

I was wondering with the new version 4 how one would go about having the ability to expose the export chart capabilities?

The way I've previously done it is via declaring the HighchartsChartContext.Consumer and extracting the chart object from there.
Then I can use the export chart function

The reason I'm doing this is I don't want the out of the box export burger button and I want to trigger the export capabilities in my own custom button.

Thanks!

anajavi:
> The way I've previously done it is via declaring the HighchartsChartContext.Consumer and extracting the chart object from there.

You can access the chart object with useChart hook:
```jsx
const ChartExporter = () => {
  const chart = useChart();
  // do things with chart

  return null;
}

<HighchartsChart>
  <ChartExporter />
</HighchartsChart>
```

filoromz:
thanks @anajavi , that worked like a charm!
I was able to call the exportChart function by being able to have reference the chart object
```
const ChartExporter = (props: any) => {
  const chart = useChart();

  props.chartRef.current = chart.object

  return null;
}
```

and then whenever i need or want to export the chart I can just do this
```
chartRef.current.exportChart();
```

anajavi:
Might be nice to provide chartRef with ref too, like this (doesn't work currently):
```jsx
const chartRef = useRef();

<Chart ref={chartRef} />
```

Shouldn't be too hard to implement. https://reactjs.org/docs/forwarding-refs.html

Praveen8529:
Hi @filoromz / @anajavi, can we have basic example for custom export for highchart using highstock..?

anajavi:
> Hi @filoromz / @anajavi, can we have basic example for custom export for highchart using highstock..?

@Praveen8529 I have never done any kind of export, so I might be wrong person to make example :)

filoromz:
hi @Praveen8529 .. im not too familiar with highstock, but i'm sure you can do something similar to the example provided above.
As @anajavi suggested, there's a hook called "useChart()" which gives you access to the underlying chart object.
[...]

PR #303, opened 2020-08-19
Add typescript definitions

anajavi:
### what?
This adds typescript definition for the library. Not complete, but all series types should have types.

References highcharts types.

Somewhat related to discussion in #113 and in https://github.com/whawker/react-jsx-highcharts/issues/267#issuecomment-578111704

### why?
- Better IDE experience even when writing plain js. 
- Type checking for those who use typescript

### caveats

- Event handlers are not typed, instead the evented components accept any props in addition to highcharts options

I am no way expert in writing definition files. I suggest that we release this and see what happens? It's easy to remove if it causes too much a headache. Maybe adding something in the release notes about "beta types".


veloek:
We've been maintaining our own typings for (a subset of) RJH, so this is really great!

You haven't considered moving the codebase towards typescript? I realise it's quite some work to get there, but I think it would be worth while. If not, please make sure to keep the typings up-to-date. Misleading types are worse than no types.

anajavi:
> We've been maintaining our own typings for (a subset of) RJH, so this is really great!

Can you please paste this definition to your module declaration and report if it works for you?

> You haven't considered moving the codebase towards typescript? I realise it's quite some work to get there, but I think it would be worth while.

That's up to whawker to decide. Previously typescript was quite cumbersome to use with react, but latest releases have been better. It's however quite an effort and adds complexity.

veloek:
> Can you please paste this definition to your module declaration and report if it works for you?

We're using the react-jsx-highstock package and the highstock specific components are missing, but other than that it seems to work. There are a few `any` types though, which could be improved on further.

anajavi:
> There are a few `any` types though, which could be improved on further.

Do you happen to have definitions for those?

veloek:
Well, I would have to go through the components and look at the code. One simple example is `Credits` where the props type should be `{enabled: boolean}`.

anajavi:
> Well, I would have to go through the components and look at the code. One simple example is `Credits` where the props type should be `Highcharts.CreditsOptions & { enabled?: boolean }`.

I added type for ```<Credits />``` in 19a4c688e9ee87a0b9852bebcab3c53e769deef6

veloek:
Nice @anajavi, this is looking good! I would like to have the event handlers listed out though, instead of `[x: string]: any`, so they also get their appropriate types.

anajavi:
> Nice @anajavi, this is looking good! I would like to have the event handlers listed out though, instead of `[x: string]: any`, so they also get their appropriate types.

Something like this should probably work:
```ts
  type ChartProps = {
    onSelection?: Highcharts.ChartSelectionCallbackFunction;
    [x: string]: any; // TODO: this is here to allow eventhandlers like onAfterAddSeries
  } & Partial<Highcharts.ChartOptions>;

  export function Chart(props: ChartProps): ReactElement;
```

There are however more events in highcharts than there are in their typings. One example is ```afterAddSeries```.

anajavi:
> Nice @anajavi, this is looking good! I would like to have the event handlers listed out though, instead of `[x: string]: any`, so they also get their appropriate types.

@veloek I added the event handler typings if you like to try.

whawker:
> You haven't considered moving the codebase towards typescript?

To be honest I don't know Typescript, so I would struggle to maintain the project. It is something I intend to learn but I don't use it day-to-day currently, like I do with JS

> I suggest that we release this and see what happens?

Worth releasing as a beta first?

anajavi:
> Worth releasing as a beta first?

This certainly won't break anybodys application, but might confuse a bit. Maybe a beta release and mention in the #259 might get some feedback?

I intend to add some kind of test or eslint configuration for this later so we see if new highcharts release breaks the typings(most likely the eventhandlers).

whawker (approved):

whawker:
Apologies for the insane delay on this, life during Covid just ends up trying to keep way too many balls in the air at once.

I've published a beta `react-jsx-highcharts@4.2.0-beta.1` containing the types added here.

Would those of you interested (@anajavi, @veloek) mind giving them a whirl?

anajavi:
Works nicely at my end.

PR #304, opened 2020-08-21
Update README to point to HighchartsProvider instead of withHighcharts

anajavi:
This removes mentions of withHighcharts in the READMEs. Also adds "DEPRECATED" to react-jsx-highstock-datepickers.

I think the react-jsx-highstock-datepickers should also be deprecated in npm registry with:
```sh
npm deprecate react-jsx-highstock-datepickers "<message>"
```
https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

### DIFF
--- a/packages/react-jsx-highcharts/README.md
+++ b/packages/react-jsx-highcharts/README.md
@@ -109,8 +109,8 @@ title: {
 ## Example
 
 ```jsx
-render () {
-  return (
+const MyChart = (props) => (
+  <HighchartsProvider Highcharts={Highcharts}>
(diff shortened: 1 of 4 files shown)


whawker (approved):
Yep, this aligns with the docusauras work I've been working on - with the editable examples like those in React Leaflet

anajavi:
> Yep, this aligns with the docusauras work I've been working on - with the editable examples like those in React Leaflet

That's nice! I have been checking if docusaurus v2 stabilises but so far it has seen quite a lot of breaking changes.

PR #310, opened 2020-09-02
Fix effect cleanups for react 17

anajavi:
### what?
[In React 17 the effect cleanup functions run asynchronously.](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#effect-cleanup-timing).

This PR changes the cleanup functions so, that they should work in React 17 too.

### why?

React 17 should release this week.


whawker (approved):
Great spot on these.

anajavi:
> Great spot on these.

We might have to make unmounting tests async too, but let's see when enzyme lands react 17 support.

PR #314, opened 2020-10-12
drop lodash-es dependency from react-jsx-highcharts

anajavi:
### what?

This replaces ```attempt(...)``` calls with ```try {...} catch { // ignore }```

### why?

It allows us to drop lodash-es dependency from react-jsx-highcharts, other libs still use some lodash functions.

This reduces bundle size by some 1kB.


whawker:
Looks good, although I think we need to remove the `LodashWebpackPlugin` from [`webpack.config.js`](https://github.com/whawker/react-jsx-highcharts/blob/0689aebecea2cf36150b258fbfd722010d334e22/packages/react-jsx-highcharts/webpack.config.js#L3)?

Would you mind removing `lodash-es` from the Highstock/Highmaps `package.json`s too?

anajavi:
> Looks good, although I think we need to remove the `LodashWebpackPlugin` from [`webpack.config.js`](https://github.com/whawker/react-jsx-highcharts/blob/0689aebecea2cf36150b258fbfd722010d334e22/packages/react-jsx-highcharts/webpack.config.js#L3)?

It is removed from  webpack.config.js

> Would you mind removing `lodash-es` from the Highstock/Highmaps `package.json`s too?

Highstock and Highmaps still have a couple of imports from lodash-es, so they need some extra work still. Maybe we merge this and remove them in a separate PR?

whawker:
> It is removed from webpack.config.js

🤦 my apologies

> Highstock and Highmaps still have a couple of imports from lodash-es, so they need some extra work still. Maybe we merge this and remove them in a separate PR?

I clearly haven't woken up this morning, good spot

whawker (approved):

whawker:
I'll push out a new release

PR #315, opened 2020-10-22
Allow react 17 as peerDependency

anajavi:
Allow react 17 as peerDependency.

The enzyme react 17 support is not yet released: https://github.com/enzymejs/enzyme/issues/2429. I don't know if we should wait for enzyme, it might take a while though?

According to release notes seems quite safe: https://reactjs.org/blog/2020/10/20/react-v17.html


whawker:
Yep, seems to work fine as far as I can see. Lets push this out.

whawker (approved):

Issue #308, opened 2020-08-31
onSetExtremes prop is not updated


hukka:
#### Bug or Feature Request?
Bug.

#### Description
Trying to update the onSetExtremes event handler (for closures) will not work, but the initial value is always used. This can be surprising when trying to close over the previous extremes in the event handler (for example to check if new data needs to be loaded). It can be worked around with refs, but this should be documented.

#### Live demo demonstrating bug
https://stackblitz.com/edit/base-chart-example-closure-in-event-handler?file=MyChart.js
open the console, change extremes using the range selectors. Notice how the logged value in the chart's render function keeps changing, but the closed value logged from the event handler is always the same (the first one), even though a new event handler is created on every render.

hukka:
Perhaps a matter of using `.update` vs `.addEvent`, as mentioned in https://github.com/highcharts/highcharts/issues/6538 ?

anajavi:
The axis events are bound only on createAxis:
https://github.com/whawker/react-jsx-highcharts/blob/c3d3106d42195415e6afa6475ed9ad8821feae97/packages/react-jsx-highcharts/src/components/Axis/Axis.js#L65

This is probably a bug or missing functionality. @whawker is there are reason why the axis eventhandlers are not updated when prop changes?

As there are more problems with highcharts event handlers (for example: https://github.com/highcharts/highcharts/issues/12067), I have been thinking about proxying all the eventhandlers so that they don't need to be passed to highcharts at all.

Something like this:
```js
useCallback((eventName, ...args) => {
  props.[`on${eventName}`](...args);
},[]);
```

It is however a little tricky as some of those events need to pass ```this``` bound to something for example [legendItemClick](https://api.highcharts.com/class-reference/Highcharts#.SeriesLegendItemClickCallbackFunction)

whawker:
> @whawker is there are reason why the axis eventhandlers are not updated when prop changes?

I never really envisioned the event handlers changing, which is an obvious oversight now 🤦‍♂️ 

If we we're to implement something similar to `useModifiedProps` for event handlers, we would need to have documentation on function equality between renders (i.e. wrapping in the event handler in `useCallback`) otherwise we might end up adding the same event handler multiple times?

Going completely off topic, I've been wondering if `useModifiedProps` is a mistake, currently when props change we figure out which ones have changed, and commit them as updates to the wrapped Highcharts chart. 

[...]

Issue #252, opened 2019-11-22
getModifiedProps does not return omitted props


anajavi:
#### Bug or Feature Request?
Bug, but probably not a regression

#### Description

getModifiedProps returns only those props that have a new value. If value is omitted from currProps, no change is detected.

Almost all components depend on getModifiedProps to handle prop changes.

#### How to reproduce
```javascript
  it('should return prop that changes to undefined', () => {
    const prevProps = {
      prop: 'testValue',
      prop2: 'testValue2'
    };

    const currProps = {
      prop2: 'testValue2'
    };

    const returnedProps = getModifiedProps(prevProps, currProps);
    expect(returnedProps).toHaveProperty('prop');
    expect(returnedProps).toEqual({ prop: undefined });
  });
```

#### Versions
* React JSX Highcharts version: 4.0.0-alpha.8
* Highcharts version: 7.2.1
* React version: 16.12.0

anajavi:
I don't think this makes a difference. I am quite sure that fixing this would make highcharts behave badly.
Closing.

PR #318, opened 2020-12-08
Allow series eventhandlers to be updated

anajavi:
This adds code to change series eventhandlers when they are changed in the props.

Somewhat related to #308.  If there are no problems with updating series eventhandlers, then it should not be too hard to use same mechanism in axis too.

I had to add dependency for react-testing-library as the tests were incredibly complex with enzyme. I am hoping we could migrate the other tests use react-testing-library as well.

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Series/Series.js
+++ b/packages/react-jsx-highcharts/src/components/Series/Series.js
@@ -76,7 +76,9 @@ const Series = memo(
     const prevProps = usePrevious(seriesProps);
 
     useEffect(() => {
-      if (!prevProps || !seriesRef.current) return;
+      if (!prevProps) return;
+      if (!seriesRef.current) return;
+
       const series = seriesRef.current;
       const { visible, data, ...rest } = seriesProps;
 
@@ -95,7 +97,23 @@ const Series = memo(
 
       const modifiedProps = getModifiedProps(prevProps, rest);
       if (modifiedProps !== false) {
-        series.update(modifiedProps, false);
+        const nonEventProps = getNonEventHandlerProps(modifiedProps);
+        series.update(nonEventProps, false);
+
+        // update changed eventhandlers
+        const modifiedEvents = getEventsConfig(modifiedProps);
+        const prevEvents = getEventsConfig(prevProps);
+        Object.keys(modifiedEvents).forEach(eventName => {
+          const oldHandler = prevEvents[eventName];
+          if (oldHandler) {
+            Highcharts.removeEvent(series, eventName, oldHandler);
+          }
+          const newHandler = modifiedEvents[eventName];
+          if (newHandler) {
+            Highcharts.addEvent(series, eventName, newHandler);
+          }
+        });
+
         doRedraw = true;
       }
       if (doRedraw) {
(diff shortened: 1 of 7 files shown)


anajavi:
I tried to do this with useManualEventHandlers also.

It however caused breaking changes, as the event handlers would not be present during creation of the series: the afterAddSeries would not be fired.

packages/react-jsx-highcharts/src/components/Series/Series.js
```diff
@@ -95,7 +97,23 @@ const Series = memo(
...
       const modifiedProps = getModifiedProps(prevProps, rest);
       if (modifiedProps !== false) {
-        series.update(modifiedProps, false);
+        const nonEventProps = getNonEventHandlerProps(modifiedProps);
+        series.update(nonEventProps, false);
+
+        // update changed eventhandlers
+        const modifiedEvents = getEventsConfig(modifiedProps);
+        const prevEvents = getEventsConfig(prevProps);
+        Object.keys(modifiedEvents).forEach(eventName => {
```
whawker, line 106: This works really well, apart from the in the case of #252. If an event handler is removed completely, the handler itself won't be removed.

Current this test case fails
```js
it('removes onHide eventhandler when prop is omitted', () => {
  const onHide = jest.fn();

  const wrapper = render(<Component onHide={onHide} />);

  wrapper.rerender(<Component />);

  seriesRef.current.object.hide();
  expect(onHide).not.toHaveBeenCalled();
});
```

Obviously that is existing broken functionality, but I'm wondering if it's something we need to revisit?

anajavi: When I tried to solve #252, I found out that some properties cannot be set to null with chart.update. They should be set to their default values when omitted from props.

I cannot remember the exact details, but for example:
- [enableMouseTracking](https://api.highcharts.com/highcharts/series.line.enableMouseTracking) default value is true, setting it to null would make it falsy.
- [tooltip](https://api.highcharts.com/highcharts/series.line.tooltip) is an object property, setting that to null can have strange side effects

So when omitting props this library would need to know what to set in every property.

I just gave up when I realised that.

I think that for now it is better to have consistently broken functionality rather than fixing it for event handlers only.

Thoughts?

anajavi:
I've converted most of the tests to use react-testing-library. Do you want me to include that change in this PR or in a separate one?

packages/react-jsx-highcharts/src/components/Series/Series.js
whawker: My only thoughts are if we can somehow utilise [`Highcharts.defaultOptions`](https://api.highcharts.com/class-reference/Highcharts#.defaultOptions), but that would likely be a breaking change due to unforeseen consequences.

I've been getting more comfortable with Typescript recently, whilst contributing to [Remirror](https://www.github.com/remirror/remirror), so thinking more and more about a Typescript rewrite, maybe something to consider then?

anajavi: Typescript can be setup for gradual upgrade too. If you want, I can do that when I have the time.

Should we merge this as is?

whawker (approved):

anajavi:
@whawker should we release a beta with this?

anajavi:
@whawker do you mind if I release a beta? I suppose it can be done with ```npm publish --tag beta```

whawker:
Apologies, I had loads of meetings yesterday and crashed out at the end of the day.

Yes, feel free to release a beta.

anajavi:
Released 4.2.1-beta.0

PR #320, opened 2020-12-09
Convert highstock tests to use react-testing-library

anajavi:
This replaces enzyme in react-jsx-highstock with react-testing-library.

Also updates react to v17.


whawker (approved):

packages/react-jsx-highstock/test/test-helper.js
```diff
@@ -1,32 +1,8 @@
...
-    .map(prop => Object.getOwnPropertyDescriptor(src, prop));
-  Object.defineProperties(target, props);
-}
-
-//const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
-//const { window } = jsdom;
-window.Date = global.Date;
-
 const nodeCrypto = require('crypto');
 window.crypto = {
```
whawker, line 2: I know I added this, but I have no idea why...

anajavi: I remember that I tried to remove it, but something broke..

PR #332, opened 2021-06-11
Destroy chart without requestAnimationFrame

### LINKED ISSUE #339: Strange behavior when RangeSelector and Navigator is used together
#### Bug or Feature Request?
Bug

#### Description
Something weird happens when you have a stock chart with both RangeSelector and Navigator. When the component is unmounted, it is moved/added directly under document.body. This happens on every unmount, creating multiple copies of the same element.

I've tested different versions of Highcharts and it seems like it was introduced with Highcharts 9.0.0 because I can't reproduce it in version 8.2.2. 

#### How to reproduce
I've made a sandboxed example that demonstrates the behavior with a comparison between native Highcharts and RJH.

https://codesandbox.io/s/react-highcharts-comparison-13eix

#### Versions
* React JSX Highcharts version: 4.2.1
* Highcharts version: 9.1.2
* React version: 17.0.2

anajavi:
### what?

Remove postponing chart.destroy with requestAnimationFrame. In react 17 the useEffect cleanup is asynchronous itself.

### why?

fixes #339

The postponing of chart.destroy with requestAnimationFrame should not be needed anymore and it might actually cause subtle bugs
I traced it's introduction to https://github.com/whawker/react-jsx-highcharts/commit/f15826cb2380641c95f9048722f210399f4c73b9

This also fixes crash with react 18 ```<StrictMode>```.


anajavi:
I've confirmed that current master crashes when running inside react@18-alpha ```<StrictMode>```.

This PR fixes the crash. Just removing requestAnimationFrame without layoutEffect changes fixes the problem too.

[React 18 runs the effect hooks twice when in StrictMode.](https://github.com/reactwg/react-18/discussions/19)
So basechart does:
- create chart
- requestAnimationFrame(chart.destroy())
- create chart

anajavi:
I've restored the cleanup inside useEffect as that is asynchronous in react >= 17.

This now only changes behaviour when used with react 16.

anajavi:
Currrent master can still crash on react 18 ```<StrictMode>```, although it is not so easy to trigger anymore.

whawker (approved):

PR #333, opened 2021-06-11
Allow axis events to be updated. fixes #308

### LINKED ISSUE #308: onSetExtremes prop is not updated

anajavi:
### what?

Allow updating axis eventhandlers.

https://github.com/highcharts/highcharts/pull/15658 fixed updating axis evenhandlers. This PR just adds code to call updates for events.

I have a feeling this might be a breaking change, but I am not sure. At least it requires highcharts 9.1.1 to work properly.

fixes #308


whawker (approved):
In combination with #335, seems like this should be a major version.

anajavi:
> In combination with #335, seems like this should be a major version.

I'd like to also drop prop-types at the same time. The typescript typings are much more complete. Is that okay?

PR #335, opened 2021-06-11
Drop support for old browsers

anajavi:
### what?

This is a breaking change along with #334

This sets browser support as follows:
- chrome, edge last 10 versions
- Firefox last Extended Support Release
- Safari last IOS version supported by Apple

 ### why?

- IE11 is essentially dead and [Microsoft has dropped support for Edge Legacy too on March 9](https://blogs.windows.com/msedgedev/2021/03/09/microsoft-edge-legacy-end-of-support/).
- It saves some space
- Dropping transpilation uncovers bugs (for example the changes to tests in this PR)
- quick google search seems to indicate that create-react-app actually transpiles node_modules with user settings anyway

This is just a suggestion, feel free to ignore it.


packages/react-jsx-highcharts/package.json
```diff
@@ -72,6 +71,10 @@
     "prop-types": "^15.0.0"
   },
   "browserslist": [
-    "ie >= 11"
+    "chrome >= 81",
+    "edge >= 81",
+    "firefox >= 78",
+    "ios >= 12",
+    "safari >= 12"
```
whawker, line 78: Might be able to replace this with something like `'since 2020'`

anajavi: The ["since 2020"](https://browserslist.dev/?q=c2luY2UgMjAyMA%3D%3D) picks up QQ Browser, which I don't know at all.

The "since 2020" browsers can also change later on to include an unknown browser. I kind of like to be explicit here.

The next react-router has also defined the browsers the same way: https://github.com/remix-run/react-router/blob/main/.browserslistrc

whawker (approved):

PR #338, opened 2021-06-16
Add typescript definitions for react-jsx-highstock

veloek:
This builds on #303 and extends types to also cover the
react-jsx-highstock package.

Please verify that the type definitions makes sense, @anajavi and and
@whawker. I used the types you defined, @anajavi, as reference and tried
to compare props of RJH components with Highcharts config types.

It works for our usage of RJH, but not all components are tested.

### DIFF
--- a/packages/react-jsx-highstock/package-lock.json
+++ b/packages/react-jsx-highstock/package-lock.json
@@ -2543,6 +2543,29 @@
       "integrity": "sha512-PijRCG/K3s3w1We6ynUKdxEc5AcuuH3NBmMDP8uvKVp6X43UY7NQlTzczakXP3DJR0F4dfNQIGjU2cUeRYs2AA==",
       "dev": true
     },
+    "@types/prop-types": {
+      "version": "15.7.3",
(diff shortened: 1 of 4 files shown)


anajavi:
Hey this is great!

I tested it a bit and it seems to work nicely.
Although there are a couple of extensions in ```node_modules/highcharts/modules/stock.d.ts ``` to base highcharts types, including them would make this a bit too complex for my taste. We can add them later if somebody needs them.

@whawker I think we could merge this as is?

veloek:
@anajavi Thanks for the quick feedback!

Regarding the extensions in `stock.d.ts`, as far as I can see those are merged in already when you import Highcharts from "highcharts/highstock".

anajavi:
@whawker do you mind if I merge this and release as an alpha?

whawker:
Yeah, go for it!

anajavi:
released as react-jsx-highstock@4.3.0-alpha.1

anajavi:
I guess there aren't any problems with this, so I am gonna release this.

PR #344, opened 2021-08-14
Added TypeScript type definitions for react-jsx-highmaps

AspieCoder1:
Added TypeScript type definitions for react-jsx-highmaps. Can someone please check that they look correct? I have tested them with the example code in the documentation and got not type errors.

### DIFF
--- a/packages/react-jsx-highmaps/types/index.d.ts
+++ b/packages/react-jsx-highmaps/types/index.d.ts
@@ -1,6 +1,5 @@
 import type * as Highcharts from 'highcharts';
 import type { ReactElement, ReactNode } from 'react';
-import * as React from 'react';
 import type {
   HighchartsChartProps,
   SeriesProps,
--- a/packages/react-jsx-highmaps/types/index.d.ts
+++ b/packages/react-jsx-highmaps/types/index.d.ts
@@ -1,0 +1,79 @@
+import type * as Highcharts from 'highcharts';
+import type { ReactElement, ReactNode } from 'react';
+import * as React from 'react';
+import type {
+  HighchartsChartProps,
+  SeriesProps,
+  AxisProps
+} from 'react-jsx-highcharts';
+
+export * from 'react-jsx-highcharts';
+
+type HighchartsMapChartsProps = HighchartsChartProps & {
+  map: Highcharts.GeoJSON;
+};
+
+export function HighchartsMapChart(
+  props: HighchartsMapChartsProps
+): ReactElement;
+
+// Series
+export function MapSeries(
+  props: SeriesProps<Highcharts.SeriesMapOptions>
+): ReactElement;
+
+export function MapPointSeries(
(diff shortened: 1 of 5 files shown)


packages/react-jsx-highmaps/types/index.d.ts
```diff
@@ -0,0 +1,78 @@
+import type * as Highcharts from 'highcharts';
+import type { ReactElement, ReactNode } from 'react';
+import type {
+  HighchartsChartProps,
+  SeriesProps,
+  AxisProps
+} from 'react-jsx-highcharts';
+
+export * from 'react-jsx-highcharts';
```
anajavi, line 9: It think this should not export everything from react-jsx-highcharts as the real index.js only exports a subset:
https://github.com/whawker/react-jsx-highcharts/blob/918b92d60bd6cc4590095f7c79126e73581ae0a9/packages/react-jsx-highmaps/src/index.js#L2-L21

anajavi:
This is great!

I am not familiar with highmaps, but I will try this out. On the surface it seems to be as it should be.

packages/react-jsx-highmaps/types/index.d.ts
AspieCoder1: I didn't see that bit in the index.js. I think the,
```js
export * from 'react-jsx-highcharts'; 
````
Should be
```js
 export { 
   Chart, 
   ColorAxis, 
   Credits, 
   Debug, 
   HighchartsProvider as HighmapsProvider, 
   Loading, 
   Legend, 
   Series, 
   Subtitle, 
   Title, 
   Tooltip, 
   useHighcharts, 
   useChart, 
   useAxis, 
   useSeries, 
   useModifiedProps, 
   withHighcharts as withHighmaps, 
   withSeriesType 
 } from 'react-jsx-highcharts'; 
```
instead. I can go and fix it so that the exports are more correct.

anajavi: I think that the useModifiedProps is exported accidentally in react-jsx-highmaps. It is for internal use only.

anajavi (changes requested):
I tried this out and it seems to work nicely when the ```export * from 'react-jsx-highcharts'; ``` is changed to what you proposed.

However the package.json still needs the same changes as in #338.
1. add types directory to files array, otherwise the types/index.d.ts does not end up in npm package
2. add "test:types": "tsc --noEmit" to scripts, this is for testing the types
3. @types/react to devDependencies, the above test fails without it

Can you do the changes above?

AspieCoder1:
I have added the changes to package.json and I have removed the export of useModifiedProps from index.js.

anajavi:
> I have added the changes to package.json and I have removed the export of useModifiedProps from index.js.

Great!

@whawker do you mind if I merge and release this as an alpha again?

whawker:
@anajavi Yep, go for it!

anajavi (approved):

PR #362, opened 2022-04-04
Fix build and dependency badges

anajavi:
This fixes the badges in README.

The build status was still referencing travis.
The david-dm.org has apparently been down for some time already.


anajavi:
Rebased this.

PR #378, opened 2022-12-05
Replace U+00A0 character with normal space

veloek:
These probably snuck in by mistake at some point 'cause I see no point of using non-breaking space in package.json.

### DIFF
--- a/packages/react-jsx-highcharts/package.json
+++ b/packages/react-jsx-highcharts/package.json
@@ -65,7 +65,7 @@
     "uuid": "^9.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


anajavi:
> These probably snuck in by mistake at some point 'cause I see no point of using non-breaking space in package.json.

You are right that these have snuck in by mistake. Did these cause problems for you? Do you think these warrant a release?

veloek:
No no, it's just a clean up 🙂  I found a few of these in our code base and when I did a full search I noticed some matches in package-lock.json so I thought I'd send a patch.

PR #380, opened 2023-04-26
Support highcharts v11

anajavi:
Seems to work as before.

The SunburstSeries doesn't have axes anymore, I changed the test accordingly.

Maybe a release after this?


whawker:
Apologies, yep go ahead!

PR #388, opened 2024-08-08
Replace highmaps shallow tests

anajavi:
continuation of #385

in order to support react 19

```<ContextSpy>``` is copied from highcharts tests

This should be last shallow-renderer replacement.


whawker (approved):
Great work on all of this!

anajavi:
> Great work on all of this!

Thanks!

anajavi:
I think react-test-renderer needs to be removed also, fortunately there is only one test suite using that.

All the other tests pass on react@19.0.0-rc.0.

PR #389, opened 2024-08-08
Replace react-test-renderer

anajavi:
This removes react-test-renderer.

I split the XAxis test to two to test highstock separately.

With this change tests should pass with react 19 too.


anajavi:
@whawker do you think this could be merged? It is not urgent though.

anajavi:
@whawker can we merge this too? react@19.0.0-rc.1 just came a couple of weeks ago.

whawker (approved):

PR #391, opened 2024-11-29
Support highcharts v12

anajavi:
Add highcharts v12 as peer dependency

One breaking change on highcharts side: modules run themselves instead of providing function to run.
For example:
```js
import 'highcharts/modules/accessibility';
```

The change does not need changes to the library, so I think we can release this as is. So maybe a release after merge?


whawker:
Might need to update the warning output in [`utils/warnings.js`](https://github.com/whawker/react-jsx-highcharts/blob/d47c17b3a79fb60b9a8efaee8589775bda6d8351/packages/react-jsx-highcharts/src/utils/warnings.js#L41).

Maybe it just outputs that "this line is not required in High charts 12 and above"?

whawker:
Whoops, sorry thought I'd replied. Yep that change is great

anajavi:
> Whoops, sorry thought I'd replied. Yep that change is great

Should we merge and release?

whawker (approved):
Looks good to me!

PR #392, opened 2024-12-08
Support react 19

anajavi:
The tests pass and the examples work.

I think it is good to go.


packages/react-jsx-highmaps/package.json
```diff
@@ -48,8 +48,8 @@
   },
   "peerDependencies": {
     "highcharts": "^9.1.2 || ^10.0.0 || ^11.0.0 || ^12.0.1",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0"
+    "react": "^17.0.0 || ^18.0.0 || ^19.0.0",
```
whawker, line 51: Optional, but could just do `17 - 19`?

whawker (approved):

packages/react-jsx-highmaps/package.json
anajavi: Yeah, it is just so little used syntax, that finding examples with google is hard.

PR #394, opened 2026-05-29
Upgrade uuid to v14

anajavi:
This upgrade is a bit more invasive, so I'd like an opinion on it.

uuid v14 is now ESM-only, and jest needs to be run with `node --experimental-vm-modules `because of it. The ESM support in jest has been that way for years and probably will remain that way for a bit longer.

I removed separate test scripts for packages and left only the top level test that tests all packages.

**Another thing: How would you feel about making react-jsx-highcharts esm only too?**
Now that node can do require(esm) from commonjs, a lot of projects are dropping support for commonjs altogether.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -6,7 +6,7 @@ import globals from 'globals';
 import reactPlugin from 'eslint-plugin-react';
 import reactHooksPlugin from 'eslint-plugin-react-hooks';
 import reactPerfPlugin from 'eslint-plugin-react-perf';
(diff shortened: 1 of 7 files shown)


whawker:
Would it be worth switching from `jest` to `vitest` instead of running in experimental mode?

anajavi:
> Would it be worth switching from `jest` to `vitest` instead of running in experimental mode?

It might be. Seems to be quite popular way nowadays. I might take a look when I find the time.

PR #396, opened 2026-08-25
Switch jest to vitest

anajavi:
Changing jest to vitest came up in #394.
Here it is.

- There should be no changes to any functionality
- Changes to tests should be just the framework changes
- To make jsx work with the vitest, the simplest way was to change extensions for all files containing jsx
- I left out the uuid package upgrade from this, but I tried it and it worked without problems

@whawker is it clean enough for merge?


whawker (approved):
Looks good to me!

anajavi:
> Looks good to me!

How do you feel about making the packages ESM only? It seems to be the trend now as node can require(esm) from commonjs modules.

PR #397, opened 2026-08-26
Add typescript tooling

anajavi:
This adds typescript tooling:
- no types are added to the source as that is much bigger task
- only small changes were made to make typescript-eslint happy
- the previous types are still used on npm release

The idea is to add typings to source in separate PRs. And maybe make a ESM only major release later.

Most of the development tooling nowadays expects typescript instead of plain javascript.

@whawker What do you think? I am entirely happy to drop this also, so don't hesitate to say if you don't like to merge it.


whawker (approved):
Awesome, this looks great!

anajavi:
> Awesome, this looks great!

Great!

PR #398, opened 2026-08-27
Fix RangeSelectorButton not unmounting

anajavi:
Found while converting react-jsx-highstock to typescript via #397.

The unmount did not actually remove the button from the rangeselector.

Typescripting old code sometimes actually finds bugs..


whawker (approved):
Nice!

PR #400, opened 2026-08-27
 Add types to react-jsx-highmaps

anajavi:
- Move types from declaration file to source code
- Switches react-jsx-highmaps to built declaration

Never used highmaps, but I tried to be careful.


anajavi:
@whawker can you take a look a this when you find the time

packages/react-jsx-highcharts/types/index.d.ts
whawker, line 192: Could we do something here like

```ts
// 1. Define the allowed key pattern (React style: on + Capitalized word)
type EventKey = `on${Capitalize<string>}`;

// 2. Use a Mapped Type to create the object structure
type EventListenerMap = {
  [K in EventKey]?: (...args: any[]) => void;
};
```

whawker (approved):
Looks good to me, however I don't use this module either. It was added by request here https://github.com/whawker/react-jsx-highcharts/issues/56 

@unofang and @idoromm, if by any chance you still happen to use this (appreciate a lot can change in 8 years) can you take a look?

packages/react-jsx-highcharts/types/index.d.ts
anajavi: It can be done with actually mapping the events to Highcharts types also.

It just is "a bit complicated". I may took a look later on.

anajavi: Here is an example of the typescript utility function to convert events:

```ts
/*
 Converts all keys of a type to event handler props prefixed with "on" and capitalized.
 example:
 type ChartEventsOptions = {
    addSeries?: ChartAddSeriesCallbackFunction | undefined;
    afterPrint?: ExportingAfterPrintCallbackFunction | undefined;
    beforePrint?: ExportingBeforePrintCallbackFunction | undefined;
    click?: ChartClickCallbackFunction | undefined;
 };
 
 type PrefixedChartEvents = MakeEvent<ChartEventsOptions>;
 // Resulting type:
 {
    onAddSeries?: ChartAddSeriesCallbackFunction | undefined;
    onAfterPrint?: ExportingAfterPrintCallbackFunction | undefined;
    onBeforePrint?: ExportingBeforePrintCallbackFunction | undefined;
    onClick?: ChartClickCallbackFunction | undefined;
 }
 */
export type MakeEvent<T> = {
  [K in keyof T as `on${Capitalize<string & K>}`]: T[K];
};

```

PR #401, opened 2026-08-28
Add types to react-jsx-highcharts

anajavi:
Moves types from declaration file to source code. Sorry about this being massive once again.

#400 needs to be merged first, as this breaks the highmaps types in the master.

Found one possible bug. Options3d sets axisLabelPosition to 'default'. The allowed values are: 'auto' | null.
Highcharts has most likely changed it almost 10 years ago. Did not fix it as this is big enough as it is.
https://github.com/whawker/react-jsx-highcharts/blob/f197e9dcacbe25f3675fb60de58ad4d0e1efc2d8/packages/react-jsx-highcharts/src/components/Options3d/Options3d.tsx#L30-L31


.github/workflows/test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: npm run test:dts
 
       - name: Test react-jsx-highstock types
-        run: npm run --workspace react-jsx-highstock test:types
+        run: npm run --workspace react-jsx-highstock --workspace react-jsx-highcharts test:types
```
whawker: This reminds me, should probably explore using pnpm instead these days. (not right now, just a note to self)

whawker (approved):

PR #404, opened 2026-09-07
Publish ESM only packages

anajavi:
@whawker I promise this is the last big change and it contains no real code changes

- modified all imports to use full file path required for esm
- removed unneeded react imports
- removed UMD build and webpack
- removed eslint-plugin-import as typescript in esm project checks the imports
- eslint-plugin-import does not work correctly with modern exports maps


whawker (approved):
Thanks for this. I appreciate this is pretty common now, but we should flag it prominently in the release notes.

anajavi:
> Thanks for this. I appreciate this is pretty common now, but we should flag it prominently in the release notes.

I agree. I will do a alpha release after merging this.

anajavi:
> > Thanks for this. I appreciate this is pretty common now, but we should flag it prominently in the release notes.
> 
> I agree. I will do a alpha release after merging this.

released the alphas

anajavi:
@whawker here is a draft of alpha release notes, please edit to your taste: https://github.com/whawker/react-jsx-highcharts/releases/tag/untagged-bc4e075a3ab3d8c29ec2

PR #406, opened 2026-09-08
Fix eventhandler types

anajavi:
- Add typescript utility ```EventsToHandlerProps``` that converts Highcharts event types to react eventhandlers
- Add back accidentally deleted Axis eventhandler types
- Remove unnecessary HighchartsChart ```[x: string]: unknown;``` as there are no eventhandlers on HighchartsChart
- Change ColorAxis to use ```EventsToHandlerProps```
- Change Series to use ```EventsToHandlerProps```
- Change Chart to use ```EventsToHandlerProps```

These problems were found in 6.0.0-alpha.0.

@whawker I think that after these fixes the v6 is good to go, what do you think?

### DIFF
--- a/packages/react-jsx-highcharts/src/components/Axis/Axis.tsx
+++ b/packages/react-jsx-highcharts/src/components/Axis/Axis.tsx
@@ -15,18 +15,29 @@ import type {
   XAxisOptions,
   YAxisOptions,
   ZAxisOptions,
-  Axis as HighchartsAxis
+  Axis as HighchartsAxis,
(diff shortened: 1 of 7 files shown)


anajavi:
~~I am starting to think that it is better to revert the EventsToHandlerProps typescript utility as it discards the jsdoc from the handlers.~~

The jsdoc is apparently missing regardless how the handlers are defined.

anajavi:
@whawker can you take a look a this when you find the time. I'd like to release v6 after merging this.

Issue #408, opened 2026-09-20
Crash when updating  to highcharts v13.1.0


anajavi:
An error is thrown: ```Cannot read properties of undefined (reading 'bind')```

Example: https://stackblitz.com/edit/react-zhp8wa34

Also tests fail when using highcharts v13.1.0

It is likely caused by this change in highcharts: https://github.com/highcharts/highcharts/pull/23887

```axis.addPlotBandOrLine``` and ```axis.removePlotBandOrLine``` internal methods were removed.

PR #410, opened 2026-09-20
Fix crash with highcharts v13.1 on react-jsx-highcharts v5

anajavi:
fixes #408

Same as #409 for the react-jsx-highcharts v6 , but backported to v5.


anajavi:
@whawker as this fixes a crash with react-jsx-highcharts v5 I would like to make a new v5 patch release

